PR #50, opened 2014-02-17
user can rearrange the elements of a form from the form screen

imilas:
and the page automatically saves the order of the elements


jegelstaff:
Actually, this one should go forward. Front side editing of the order would be very good, in conjunction with independent ordering of elements in the multipage screen. Whether this exact code is used, the concepts and some architecture are relevant.

jegelstaff:
See also: https://github.com/jegelstaff/formulize/pull/81

PR #64, opened 2014-03-03
Dev.amir.list reorder

imilas:
"As a System desinger, I would like to specify the order of elements in a list separately from how they appear in a form"


jegelstaff:
- one
- two
- three

jegelstaff:
- [] one
- [] two

PR #81, opened 2014-04-13
Reorder elementes in form screen

imilas:


jegelstaff:
Reordering elements from the front side! Wild idea. Totally not going to fit into the new setup.

imilas:
:(

jegelstaff:
See also: https://github.com/jegelstaff/formulize/pull/50

jegelstaff:
We will target 8.1 with an ability to reorder elements within a screen page as a configuration option on the admin side, and this ability to reorder the elements from the actual form as well. In conjunction with the slide panel for options when editing on the front side, perhaps that actually is the graphical form builder!! 🤔

PR #115, opened 2014-09-20
Elements with and without data.

Lastiritas:
Finish the implementation and also changed all the places where getVar('elements') needed to be replaced by getVar('elementsWithData')


jegelstaff:
There is a useful feature here to get a list of elements in a form that have data. This should be merged in, and used widely.

jegelstaff:
Not every part of this might be necessary, but the addition of that property to the form object at the very least should be included.

jegelstaff:
This has been incorporated into the PR https://github.com/jegelstaff/formulize/pull/540

PR #124, opened 2014-10-02
Fixing one to one links when no all done button

jegelstaff:
Needs test.  A regular form screen that uses a 1-1 relationship, and has the All Done button turned off, needs to be made.  And then when saving values in both forms, the value saved in the non-main form should be remembered when you go back into the form.  (Without this change, the value in the non-main form would be saved, but the linkages would not be set between the forms, and then when you go back in, the non-main form would be treated as a "new" entry rather than having a specific entry id pulled in.)


jegelstaff:
Still needs a test!!

PR #128, opened 2014-10-09
Improving subform blank synching

jegelstaff:
Now it happens in readelements right after entries are written, and does
not depend on determining the parent entry.  We can get the parent entry
based on the fields that were written, or on the submitted value from
the form.


JamesEarle:
:+1: All good. Altering the Code Mirror folder to not be found allowed the textarea to render properly and Selenium inserts the text no problem (y) All other tests remain unaffected as well.

jegelstaff:
Test will have to wait :(

PR #129, opened 2014-10-11
Adding prepop feature for subforms

jegelstaff:
This needs a test!  Test should set a prepopulation choice for a subform element, and then verify when the form loads that entries exist in the subform corresponding to each of the options in the chosen prepopulation source.

Also, this feature does not currently synch prepop subform entries with a main form entry that has not been saved yet.  This works fine on second and later pages of a multipage form, but when the mainform entry does not exist yet, we have to synch for that.  There is a comment about this in formdisplay.php.  Some further flags should be sent identifying the prepop entries, so we can flag them when they are saved, and then run a synch against the first entry saved in the main form.  This is very similar to the blank subform entry synching.

Ideally, we would delete the prepop entires if the main form never gets saved, but detecting that is hard, and orphaned subform entries are not going to be a huge problem in most cases, though it is definitely ugly.


JamesEarle:
:+1: and :-1: at the same time, received a StaleReferenceError or on the cr test. Probably a caching problem somewhere with Travis, but the test works locally and passes the push build, so I think it's ready. I'll try recommitting to trigger another test just to be sure.

PR #130, opened 2014-10-11
Making others work in blank subform entries

jegelstaff:
Initial subform blank entries, would not save other box values properly
before this.  We need to track the specific instance of an other box,
since there will be multiple copies of the same element in the multiple
subform blanks (a single subform blank would not likely have had the
problem).


JamesEarle:
:+1: Build is passing, should be ready to merge.

PR #132, opened 2014-10-11
Add an option to limit the range for date elements.

eikonos:
This pull request adds an option on date form elements to limit dates in the past and/or dates in the future. For example, a "Completed Date" form element might allow any date in the past, but no dates in the future. The date limits are validated when the user types or selects dates from the javascript date popup.


MORich-1:
When a limit is selected (either before or after today's date), you can currently save the selection without placing a value in the box, for the number of days to allow. While in practice it defaults to zero (having no real effect), It may be good to have the user made to put in a value, or else de-select the check box.

eikonos:
Zero is a valid value, meaning 'today', though it should actually be written in the box to make it clear that's the value. A good fix would be to use javascript to check if the text box is blank when the checkbox is clicked. In that case, focus should be set to the text box, zero should be entered and that text selected (so that it's easy to simply type another number).

MORich-1:
Aside from the above potential implementation, I have tested and reviewed the implementation, and both check boxes work as planned. Everything looks good for what it here! The suggested fix could always be a separate new issue. :+1:

MORich-1:
There, The fix was implemented. All is good here! :+1:

PR #134, opened 2014-10-22
Episani add support for selecting a view for a specific user group

jegelstaff:


jegelstaff:
134 and 126 merged together into 229, and then 229 merged into 134, and then 134 manually merged into master. Case Closed.

PR #138, opened 2014-11-05
Select specific columns in getData()

guohengw:
:+1: 
@jegelstaff This PR is good to review. For some lines of code, I don't know why the indentation looks weird on github, the indentation does look fine locally.


jegelstaff:
A clean approach to this long standing nice-to-have feature. But we have always held off because of the presumption in many cases that all fields are available in datasets. This could be implemented only for situations where columns are specified in the list? And we won't get in trouble for missing any? Worth keeping around as a source to consult if/when this issue becomes top of mind again.

PR #139, opened 2014-11-09
Add code to fill in default values for new radio buttons added to forms ...

mgoertzen:
...which already have entries.


jegelstaff:
This is an interesting approach, and the method on the data handler should be moved to the element class I think, so it could be used more generally/cleanly. It's an element maintenance issue, not a raw data access issue.

jegelstaff:
See also: https://github.com/jegelstaff/formulize/pull/203

jegelstaff:
Superseded by #831

Issue #127, opened 2014-10-07
Remove one of the two login forms from the page displayed if you fail to log in the first time.


eikonos:
If you fail to log in, you're taken to a slightly different version of the login page and there are two login forms and one forgot-email form. There should only be one username/password login form on that page.

jegelstaff:
Fixed in dc84795e4ecf0e43efdc047f5727155a0619bcdc

PR #140, opened 2014-11-10
remove redundant login form

guohengw:
Fix issue https://github.com/jegelstaff/formulize/issues/127


jegelstaff:
We can't hack the core system files like this. To remove the presence of two forms, it would be better to change the login block to only appear on the front/top page, and make sure that setting is in our custom SQL that gets written to the DB during the installation process.

jegelstaff:
Fixed in dc84795e4ecf0e43efdc047f5727155a0619bcdc

PR #141, opened 2014-11-14
Design Improvement

flaggada:
New fresh icons, with some change in css file, to make UX up to date.
Hope this could help. :-)
uximprovment


jegelstaff:
I really like this Francois, thanks!

I think we should incorporate this as a theme, once we have pull request 43 complete. Then people could simply choose which theme they want to use and it would be the "classic" one or this one.

Thanks for sharing, we really appreciate it.

flaggada:
Thanks Julian, that's a good idea. I'm happy to bring some help, to this amazing project :+1:

jegelstaff:
So much time passed. I'm sorry it never become a distinct theme. But version 7 does have a pretty nice theme and mobile friendly, for what it's worth.

PR #147, opened 2015-01-12
Treat typed val as new val if box loses focus early

jegelstaff:
This is an incomplete fix! It will lead to duplicates if the user types
a value that matches an existing one and then leaves the box early
before the autocomplete UI shows up (typed value will be saved as if it
were new, not picking up the id of the existing value).

Also, JS minification might need to be redone. I did it manually.


jegelstaff:
Instead of compensating for this in js, which would be hard and have timing problems, we may be compensating on the server side already. See this code in modules/formulize/include/functions.php:

if (is_string($ele) and substr($ele, 0, 9) == "newvalue:") {
            // need to add a new entry to the underlying source form if this is a link
            // need to add an option to the option list for the element list, if this is not a link. 
            // check for the value first, in case we are handling a series of quick ajax requests for new elements, in which a new value is being sent with all of them. We don't want to write the new value once per request!
            $newValue = substr($ele, 9);
            if ($element->isLinked) {
                $boxproperties = explode("#*=:*", $ele_value[2]);
                $sourceHandle = $boxproperties[1];
                $dataHandler = new formulizeDataHandler($boxproperties[0]); // 0 key is the source fid
                $newEntryId = $dataHandler->findFirstEntryWithValue($sourceHandle, $newValue); // check if this value has been written already, if so, use that ID
                if(!$newEntryId) {
                    $newEntryId = formulize_writeEntry(array($sourceHandle=>$newValue));
                } 
                $value = $newEntryId;
            } else {
                $value = $newValue;
                $element_handler = xoops_getmodulehandler('elements', 'formulize');
                if(!isset($ele_value[2][$newValue])) {
                    $ele_value[2][$newValue] = 0; // create new key in ele_value[2] for this new option, set to 0 to indicate it's not selected by default in new entries
                    $element->setVar('ele_value', $ele_value);
                    $element_handler->insert($element);
                }
            }
            break;
        }

Note that it only called formulizeWriteEntry if there is no entry id found when it looks for an entry with the current value in it.

So what we need to do is verify when you type too fast and tab away and you have typed something that matches an existing value in the autocomplete, and then save the entry, that what gets saved is a reference to the existing value and not a new copy of the value.

I suspect Selenium tests are too onerous for this particular functionality, which is a tiny fix, not a feature we want to have some real testing on. I can imagine some tests for this, but it would be very tricky to get the autocomplete behaviour right, since in some cases we have to wait for the UI and in some cases we specifically want to leave the UI early.

jegelstaff:
Testing too awkward. Server side seems to compensate. Forget it for now.

PR #158, opened 2015-01-18
Run a function after saving an entry.

eikonos:
This commit adds a user-editable PHP code block to be run after an entry is saved. This is useful when you want to update a value in another form based on information generated when this record is saved.


eikonos:
:+1:

jnip:
I've tested this functionality.

It correctly updates my database to include on_after_save column [to the table formulize_id]

I have tried writing code in After Saving to do php code and it works (runs the php code and only does it after the saving). 

The only thing that tripped me up was I first tried to add the php tag around my code - this made it not work. 

Did not notice anything wrong with the code itself.

PR #159, opened 2015-01-18
Fixes install SQL data to not have redundant login screen.

MORich-1:
With the searching assistance of Joe Nip (jnip), found the area in the
SQL which populates the Block_Module_Links table. Added an if statement
to the while loop to single out the login and ensure its value is set to
top page only (0-1). Tested on separate installation.


eikonos:
:+1:

PR #161, opened 2015-01-18
Added php editor for site

jujinfu:


eikonos:
Include the custom php code file from modules/formulize/initialize.php, but since there can be one code file for each Formulize Application, loop through the applications that are in the database and include the file for each of them.

jujinfu:
When we loop through all custom code, do we access the database to get all codes or do include directly from temp folder? 

And, from what we have now, each application will only have one file containing custom code? Do we allow then to create more? Or it's suppose to be only one file only containing all custom code?

PR #168, opened 2015-01-27
Update relationship.php

MORich-1:
Updated relationship.php to allow user to view arrays regarding
relationships using newer versions of PHP.


eikonos:
:+1:

eikonos:
I found a minor issue with the current code. If you create a second list screen (which uses the same relationship as the first), then the message beside the data-entry screen will be displayed twice. It should only display once.

MORich-1:
Ahh yes, I thought I had tested multiple forms and it only showed up the once, my apologies! Ill have it fixed this evening.

MORich-1:
There, I believe that after my application of the latest commit (Boolean) things will progress smoothly. I tried with up to 3 list screens, and only one warning popped up.

:+1:

PR #169, opened 2015-01-28
New graph screen

jujinfu:


modules/formulize/class/graphScreen.php
```diff
@@ -0,0 +1,128 @@
...
+			"height" => $screen->getVar('height'),
+			"orientation" => $screen->getVar('orientation'),
+			"backgroundcolor" => $bgc,
+			"barcolor" => $barc,
+			"defaultview" => $screen->getVar('defaultview'),
+			"limitviews" => $screen->getVar('limitviews'),
+			"usecurrentviewlist" => $screen->getVar('usecurrentviewlist')
+		);
+		include_once XOOPS_ROOT_PATH."/modules/formulize/include/graphdisplay.php";
+		displayGraph('Bar', $screen->getVar('fid'), $screen->getVar('frid'), $screen->getVar('labelelem'), $screen->getVar('dataelem'), $screen->getVar('ops'), $options);
```
eikonos: There's a problem here. These two variables are integers, which represent element IDs: $screen->getVar('labelelem'), $screen->getVar('dataelem'). Later in graphdislay.php, these are used like this: display($entry, $dataElement);. It works if the element has the same ID and handle, but if the element handle has been changed to text, such as, 'client_name' then it fails. The code in one of these places should be changed so that it's consistently using element IDs or handles, and it's probably better to use handles since that's what a lot of the other code does.

jujinfu:
at the beginning, all graphs are stored with label element id and data element id. And later it should be using element id any where else. I didn't notice this before. should I change all to handles? that means change from very beginning. or i just fix this by using id in other places?

jujinfu:
so far i have changed a lot in graphdisplay.php. now we will process all data then according to the graph type to generate the graph. still need some small improvements

PR #179, opened 2015-02-09
Language constant swapped for string literal.

susannahkirby:


susannahkirby:
This is a tiny one.

::+1::

modules/formulize/templates/admin/form_permissions.html
```diff
@@ -21,7 +21,7 @@
                     <{/foreach}>
                 </select>
 								<br /><br />
-								<input type="button" name="showperms" value="Show permissions for these groups" />
+								<input type="button" name="showperms" value="<{$smarty.const._AM_PERMISSIONS_SHOW_PERMS_FOR_GROUPS}>" />
```
eikonos: I searched all the files for the _AM_PERMISSIONS_SHOW_PERMS_FOR_GROUPS constant, but it might only exist in the multiple permissions branch.

susannahkirby: Thanks for catching that - I added it to the right file.

PR #180, opened 2015-02-09
Fix for making autocomplete select boxes not accept 'none' as a submitta...

MORich-1:
...ble value.

Edited autocomplete JavaScript line for case 'select', and included a
comment regarding the update.


MORich-1:
:+1:

PR #184, opened 2015-02-18
Cloned forms now carry over elements but get new form and list screens.

susannahkirby:


susannahkirby:
Still to-do:
Cloned forms should carry correct application id value
Refactor and clean-up

PR #187, opened 2015-02-23
Fixed delimiter, tested fix

MORich-1:
The issue was  quite simple here: The problem arises when the initial entry was stored with a different prefix than the one that it looked for / imploded back onto the entry inside the changeUserSubmittedValues() function. Once I had checked phpMyAdmin to see the original save prefix, I changed it and everything worked fine.

Entries Before:initialselect

Entries after using Resynch:
entrieschanged

:+1:

### DIFF
--- a/modules/formulize/class/data.php
+++ b/modules/formulize/class/data.php
@@ -983,7 +983,7 @@ class formulizeDataHandler  {
 				}
 				break;
 		}
-		$prefix = ($ele_type == "check" OR ($ele_type == "select" AND $ele_value[1])) ? "#*=:*" : ""; // multiple selection possible? if so, setup prefix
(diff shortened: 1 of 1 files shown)


modules/formulize/class/data.php
```diff
@@ -983,7 +983,7 @@ function changeUserSubmittedValues($element_id_or_handle, $newValues) {
 				}
 				break;
 		}
-		$prefix = ($ele_type == "check" OR ($ele_type == "select" AND $ele_value[1])) ? "#*=:*" : ""; // multiple selection possible? if so, setup prefix
```
eikonos: Actually, it's a little more complicated than this. One element type uses one delimiter and another element type uses the other delimiter. They should probably both use the same delimiter, but then a whole bunch of data in existing systems needs to be converted, and that might break something. It would also mean that the delimiter string through all the code needs to be changed. At this point, it's probably just easier to detect the element type here and use the appropriate delimiter string.

MORich-1:
:+1:

PR #188, opened 2015-02-26
Use date element default in subforms.

eikonos:
Now the default for date elements will be applied when used in a subform.

:+1:


jujinfu:
Tested and reviewed. It works. :+1:

PR #189, opened 2015-02-26
When cloning forms append tag

MORich-1:
Form names are now showing up in the structure: 'FormName [Cloned]' (or 'FormName [Cloned]1,2,etc.'). Only two small changes were required for this issue.

twotests


MORich-1:
:+1:

PR #191, opened 2015-03-01
Fix redirect loop

jujinfu:
When a user  trying to visit a form without permission, this will show
an error msg and put user back to it's profile page.

### DIFF
--- a/user.php
+++ b/user.php
@@ -136,7 +136,7 @@ switch ($op) {
 		exit();
 		break;
 
-	case $op == 'logout':
+	case 'logout':
 		$sessHandler = icms::$session;
 		$sessHandler->sessionClose(icms::$user->getVar('uid'));
 		redirect_header(ICMS_URL . '/index.php', 3, _US_LOGGEDOUT . '<br />' . _US_THANKYOUFORVISIT);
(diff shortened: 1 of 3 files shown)


user.php
```diff
@@ -217,4 +217,27 @@
...
+	 *Added By Jinfu Feb 2015
+	 */
+	case 'switchuser':
+		if(!icms::$user){
+			redirect_header('index.php',5,_US_NOPERMISS);
+		} elseif (!empty($_GET['xoops_redirect'])) {
+			$redirect = htmlspecialchars(trim($_GET['xoops_redirect']));
+			$isExternal = FALSE;
+			$tragetURL=substr(ICMS_URL,0,strpos(ICMS_URL,"/formulize")).$redirect;
+			redirect_header("user.php?op=main",5, "Visiting: '".$tragetURL."'<br />"._US_NOPERMISS."<br />You may switch user and try again.");
```
eikonos: In most cases, users only have one account, so if they can't view a form, they can't switch users. The message should say something like "You do not have permission to view this page."

jujinfu: should i redirect to user.php?op=main or userinfor.php?uid=.........

jujinfu:
:+1:

user.php
```diff
@@ -217,4 +217,29 @@
...
+	 *Added By Jinfu Feb 2015
+	 */
+	case 'nopermission':
+		if(!icms::$user){
+			redirect_header('index.php',5,_US_NOPERMISS);
+		} elseif (!empty($_GET['xoops_redirect'])) {
+			$redirect = htmlspecialchars(trim($_GET['xoops_redirect']));
+			$isExternal = FALSE;
+			$tragetURL=substr(ICMS_URL,0,strpos(ICMS_URL,"/formulize")).$redirect;
+			redirect_header(ICMS_URL."/modules/formulize/application.php?id=all",5, _US_NOPERMISS."<br />Viewing: '".$tragetURL."'<br />");
```
eikonos: The error message would be better if it said, "Sorry, you do not have permission to view this page. Please contact your site administrator." Adding the link to the page is not necessary.

jujinfu:
:+1:

PR #193, opened 2015-03-02
modify-buttons-on-url_accessed-data_entry

jnip:
Fixed bug. Buttons will now work as expected.

### DIFF
--- a/modules/formulize/include/formdisplay.php
+++ b/modules/formulize/include/formdisplay.php
@@ -401,11 +401,10 @@ function displayForm($formframe, $entry="", $mainform="", $done_dest="", $button
 	 * Display list of entries screen on-click of form buttons "Save and Leave" and "Leave Page".
 	 */
 	if (isset($_GET['sid'])) {
-	    $screen = xoops_getmodulehandler('screen', 'formulize')->get($_GET['sid']);
-	    if ($screen->getVar('type') == 'form') {
-		$currentURL = $_SERVER['PHP_SELF'] . "?fid=" . $screen->form_id();
+	    $curr_screen = xoops_getmodulehandler('screen', 'formulize')->get($_GET['sid']);
+	    if ($curr_screen->getVar('type') == 'form') {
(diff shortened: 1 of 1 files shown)


jnip:
:+1:

jnip:
:+1:

modules/formulize/include/formdisplay.php
```diff
@@ -395,8 +395,13 @@ function displayForm($formframe, $entry="", $mainform="", $done_dest="", $button
 
 	$mid = getFormulizeModId();
 
-	$currentURL = getCurrentURL();
-
+	$currentURL = ( (isset($_GET['ve']) && isset($_GET['fid'])) ? $_SERVER['PHP_SELF'] . "?fid=" . $_GET['fid'] : getCurrentURL() );
+	if (isset($_GET['sid'])) {
+	    list($fid) = icms::$xoopsDB->fetchRow(icms::$xoopsDB->query("SELECT fid FROM " . icms::$xoopsDB->prefix('formulize_screen') . " WHERE sid=" . $_GET['sid']));
```
eikonos: There should not be an SQL query written in this file. All SQL queries that deal with forms and screens should be in the form or screen object class. A better way to do this is to load the screen object with the right sid value, then access the fid property.

modules/formulize/include/formdisplay.php
```diff
@@ -395,8 +395,13 @@ function displayForm($formframe, $entry="", $mainform="", $done_dest="", $button
 
 	$mid = getFormulizeModId();
 
-	$currentURL = getCurrentURL();
-
+	$currentURL = ( (isset($_GET['ve']) && isset($_GET['fid'])) ? $_SERVER['PHP_SELF'] . "?fid=" . $_GET['fid'] : getCurrentURL() );
```
eikonos: There's an issue with this approach. If you have a form with two list screens, and you open the non-default list screen and edit an entry, when you click the Save And Leave or the Leave button, instead of returning to the list screen you were viewing, you will now be directed to the other list screen because it is the default.

jnip:
:+1:

PR #194, opened 2015-03-02
Clone forms refactor

susannahkirby:


eikonos:
This works for me. I cloned a couple forms, and the new forms appear, along with screens. Is there an issue when you clone a form?

susannahkirby:
It does work for me as pushed. Did you try switching which lines are commented out? When I make the call to 
> $listScreenHandler->setDefaultListScreenVars($screen, $defaultFormScreenId, $newtitle, $newfid);
(line 1265 in forms.php), I get the error.

eikonos:
The setDefaultListScreenVars() function exists in formScreen.php, but that there isn't a corresponding function with that name in the listOfEntriesScreen.php file.

susannahkirby:
Oh, of course! I see - I had that method in formScreen.php and not listOfEntriesScreen.php. Thanks, it's working now.

PR #195, opened 2015-03-04
Prevent adding new options when an autocomplete box displays a list of user names

jujinfu:
Before saving to database, autocomplete element option will be checked to make sure if there is {USERNAMES} or {FULLNAMES} selected, option will not allow new entries to be added.

also in element_optionlist.html , added javascript to make sure user doesn't break this rule.

### DIFF
--- a/modules/formulize/admin/save/element_options_save.php
+++ b/modules/formulize/admin/save/element_options_save.php
@@ -46,6 +46,7 @@ $element = $element_handler->get($ele_id);
 $ele_type = $element->getVar('ele_type');
 $fid = $element->getVar('id_form');
 
+
 $form_handler = xoops_getmodulehandler('forms', 'formulize');
 $formObject = $form_handler->get($fid);
 if($formObject->getVar('lockedform')) {
@@ -127,9 +128,9 @@ if($ele_type == "select") {
 
     global $xoopsDB;
     $sql_link = "SELECT ele_caption, id_form, ele_handle FROM " . $xoopsDB->prefix("formulize") . " WHERE ele_id = " . intval($_POST['formlink']);
-		$res_link = $xoopsDB->query($sql_link);
-		$array_link = $xoopsDB->fetchArray($res_link);
-		$processedValues['elements']['ele_value'][2] = $array_link['id_form'] . "#*=:*" . $array_link['ele_handle'];
+    $res_link = $xoopsDB->query($sql_link);
+    $array_link = $xoopsDB->fetchArray($res_link);
+    $processedValues['elements']['ele_value'][2] = $array_link['id_form'] . "#*=:*" . $array_link['ele_handle'];
   } else {
     // a user requests to unlink the select box and select box is currently linked
     if ($_POST['formlink'] == "none" AND $element->isLinked){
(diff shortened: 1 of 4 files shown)


modules/formulize/templates/admin/element_optionlist.html
```diff
@@ -54,5 +61,50 @@
     $("[name=cleardef]").blur();
   });
   <{/if}>
+  
+  <{if $content.type eq "select"}>
+  var i=0;
+  var length=<{$content.useroptions|@count}>;
+  for(i=0;i<length;i++){
+    if (($("[name=ele_value["+i+"]]").val()=="{USERNAMES}" ||$("[name=ele_value["+i+"]]").val()=="{FULLNAMES}")&&$("[name=defaultoption["+i+"]]").is(":checked")){
```
eikonos: It is good that the option to allow adding new entries is disabled, but instead of this logic here, I would have made a change in the code that saves the element settings into the database. When saving the settings, if one of the names options is selected then automatically switch to not allowing users to add new options.

eikonos:
The title and description of this pull request could describe what it does more clearly. For example, it could be something like "Prevent adding new options when an autocomplete box displays a list of user names."

jujinfu:
:+1:

eikonos:
#195

modules/formulize/templates/admin/element_optionlist.html
```diff
@@ -54,5 +63,65 @@
...
+   */
+  function checkNAMES(){
+    var x=0,y=-1,z=-1;    //y for {FULLNAMES}  z for {USERNAMES}
+    for(x=0;x<length;x++){
+      if ($("[name=ele_value["+x+"]]").val()=="{USERNAMES}"){
+        z=x;
+        if ($("[name=defaultoption["+x+"]]").is(":checked")){
+          disableForAutocomplete();
+        }
+        alert("found username"+x);
```
eikonos: Should not display an alert.

modules/formulize/templates/admin/element_optionlist.html
```diff
@@ -54,5 +63,65 @@
...
+      }
+      if($("[name=ele_value["+x+"]]").val()=="{FULLNAMES}"){
+        y=x;     
+        if ($("[name=defaultoption["+x+"]]").is(":checked")) {
+          disableForAutocomplete();
+        }
+        alert("found fullname"+x);
```
eikonos: Should not display an alert.

modules/formulize/admin/save/element_options_save.php
```diff
@@ -232,23 +233,30 @@
...
+  }
+  if(!$changeResult = $data_handler->changeUserSubmittedValues($ele_id, $newValues)) {
+    print "Error updating user submitted values for the options in element $ele_id";
+  }
+}
+
+//newly added for autocomplete box to make sure when {USERNAMES} and {FULLNAMES} are selected, system will not allow new entries to be added
+//ele_value[8] ==1 will make sure it's an autocomplete box
+//Added by Jinfu MAR 2015
+if($processedValues['elements']['ele_value'][8]==1&&($processedValues['elements']['ele_value'][2]['{USERNAMES}']==1||$processedValues['elements']['ele_value'][2]['{FULLNAMES}']==1)){
```
eikonos: This line is hard to read because it is long and there are no spaces around the == and || operators. Also, it didn't work until I changed ['{USERNAMES}']==1 to ['{USERNAMES}']==0. Does it work with 1 for you?

modules/formulize/admin/save/element_options_save.php
```diff
@@ -232,23 +233,30 @@
...
+      $newValues = $processedValues['elements']['ele_value'][2];
+      break;
+  }
+  if(!$changeResult = $data_handler->changeUserSubmittedValues($ele_id, $newValues)) {
+    print "Error updating user submitted values for the options in element $ele_id";
+  }
+}
+
+//newly added for autocomplete box to make sure when {USERNAMES} and {FULLNAMES} are selected, system will not allow new entries to be added
+//ele_value[8] ==1 will make sure it's an autocomplete box
```
eikonos: This is a very good comment, because without it, someone reading the code would not know what ele_value[8] ==1 means.

modules/formulize/templates/admin/element_optionlist.html
```diff
@@ -54,5 +63,65 @@
...
   <{/if}>
+  
+  <{if $content.type eq "select"}>
+  checkNAMES();
+  /**
+   *this function will consantly check for {USERNAMES} and {FULLNAMES}
+   */
+  function checkNAMES(){
+    var x=0,y=-1,z=-1;    //y for {FULLNAMES}  z for {USERNAMES}
+    for(x=0;x<length;x++){
```
eikonos: A loop is not needed here, because {USERNAMES} or {FULLNAMES} must be the first and only option.

jujinfu: what i was doing here is to check all the options. Like they have {USERBANES} in first box, then they have "asd" in second box, and there is another {FULLNAMES} in third box.  i need to loop through all the possible boxes to make sure either {FULLNAMES} and/or  {USERNANES}  are/is checked. 

Do you want me to make sure if they got a  {USERNANES}  or {FULLNAMES} then everything else is discarded?
Or if they got  {USERNANES}  or {FULLNAMES} selected, we automatically uncheck everything else but keep the rest?
To me, make sure one of those as first option seems unreasonable, they might have other options but not checked.
But, {USERBANES}  or {FULLNAMES} is only option checked  sounds good.

modules/formulize/admin/save/element_options_save.php
jujinfu: yes, it works on my side. you can check the new comments in the latest commit

modules/formulize/templates/admin/element_optionlist.html
eikonos: The way the {USERNAMES}/{FULLNAMES} feature works is that one of those strings has to appear in the very first option, and there are no other options at all. There can't be other items in the list. That's why those options have to appear in the very first text box.

jujinfu: OK, I'll try to alert user when they enter {USERNAMES} or {FULLNAMES}.

The msg will be something like "if you want to keep one of these, everything else will be deleted. and the very first box will be come one of these. Do you want to apply this change?"  If user click yes, it will apply the change stated above, else it will delete either {U..} or {F..} in the box. 

Does this sounds good to you ?

eikonos: No, I don't think an alert is necessary. There is help text there that says, "Setting a single option of '{FULLNAMES}' or '{USERNAMES}' will produce a list of users based on the group limits set below." and this should make it clear that those special options only work when they are the only items set in the list. As long as the option above which prevents adding new options is enforced, then it will be fine.

jujinfu:
:+1:

modules/formulize/templates/admin/element_optionlist.html
```diff
@@ -35,11 +41,12 @@
 </fieldset>
 
 <script type="text/javascript">
+  
   $("[name=addoption]").click(function (){
     number = $(".useroptions:last").attr('name');
     number = parseInt(number) + 1;
     appendContents1 = '<{if $content.type == "radio"}><input type="radio" name="defaultoption" value='+number+'></input><{else}><input type="checkbox" name="defaultoption['+number+']" value='+number+'></input><{/if}>';
-    appendContents2 = '<input type="text" name="ele_value['+number+']"></input>';
+    appendContents2 = '<input type="text" name="ele_value['+number+']" onchange="checkNAMES()"></input>';
```
eikonos: Is there a function called checkNAMES()? I don't see it anywhere in the code.

modules/formulize/templates/admin/element_optionlist.html
```diff
@@ -54,5 +61,34 @@
...
+
+  function disableForAutocomplete(){
+    $("input[name=elements-ele_value[16]]:eq(1)").attr("disabled","disabled");
+    $("input[name=elements-ele_value[16]]:eq(0)").attr("checked","checked");
+    $("[name=addoption]").attr("disabled","disabled");
+  }
+  
+  function enableForAutocomplete(){
+    $("input[name=elements-ele_value[16]]:eq(1)").attr("disabled",0);
+    /*if($("input[name=elements-ele_value[16]]:eq(0)").is(":checked")) {
```
eikonos: This code can be deleted.

jujinfu:
:+1:

modules/formulize/templates/admin/element_optionlist.html
```diff
@@ -35,11 +41,12 @@
 </fieldset>
 
 <script type="text/javascript">
+  
   $("[name=addoption]").click(function (){
     number = $(".useroptions:last").attr('name');
     number = parseInt(number) + 1;
     appendContents1 = '<{if $content.type == "radio"}><input type="radio" name="defaultoption" value='+number+'></input><{else}><input type="checkbox" name="defaultoption['+number+']" value='+number+'></input><{/if}>';
-    appendContents2 = '<input type="text" name="ele_value['+number+']"></input>';
+    appendContents2 = '<input type="text" name="ele_value['+number+']" onchange="checkForNAMES()"></input>';
```
eikonos: The actual function is checkForNames() without all-caps on NAMES, so this results in a javascript warning because the function can't be found. However, only the first text box needs to be changed anyway, so this javascript event can be removed anyway because it's not needed.

PR #196, opened 2015-03-09
Alter CSS to prevent Overlapping

MORich-1:
The formulize-admin.css file contained one div (homepage-description)
which was to be limited in its width. The other div which contained the
toolbar, had no name or space in the stylesheet, as it was hard-coded
into ui-accordion.html. I added a width limit to this as well, and now
the elements do not overlap (see below image).

fixedoverlap

### DIFF
--- a/modules/formulize/templates/admin/ui-accordion.html
+++ b/modules/formulize/templates/admin/ui-accordion.html
@@ -35,7 +35,7 @@
         <h3><a href="#"><span class="accordion-name"><{$section.name}></span></a></h3>
 		<div class="accordion-content content">
 			<{if $section.header}>
-			<div style="position: absolute; top: 10px; right: 10px;">
+			<div style="position: absolute; max-width: 50%; top: 10px; right: 10px;">
 				<{$section.header}>
 			</div>
 			<{/if}>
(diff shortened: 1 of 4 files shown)


MORich-1:
:+1:

modules/formulize/templates/admin/ui-accordion.html
```diff
@@ -35,7 +35,7 @@
         <h3><a href="#"><span class="accordion-name"><{$section.name}></span></a></h3>
 		<div class="accordion-content content">
 			<{if $section.header}>
-			<div style="position: absolute; top: 10px; right: 10px;">
+			<div style="position: absolute; max-width: 50%; top: 10px; right: 10px;">
```
eikonos: Instead of having each of these side-by-side covering half of the area, I'd prefer to have them full-width and the toolbar above the description. Can you please change it? Also, move the "Add a Form" link from the left side into the group of links in the toolbar.

MORich-1: Certainly!

MORich-1:
Here is an image of the new layout. I have also imposed a minimum width of the window canvas (measured) to prevent any other overlapping. let me know what you think.
newfix

MORich-1:
:+1:

eikonos:
The screenshot looks great! I think it is important that the toolbar options are all in line, as they are now, because it's consistent and easier to read. I will load up the code and check it out in a bit.

PR #199, opened 2015-03-13
Button Implementation for Deleting Temporary files.

MORich-1:
A New script was made in the Formulize root folder called
'deleteTempCache.php', which runs a small script to delete the cached
template files in a for each loop.
The button was made in home.php to be in line with the right-hand menu.
In ui,.html, the JavaScript was appended to include a click function for
the above button, which (only when confirmed), forms an AJAX query for
the newly created script, which is executed to delete cached files.
The min-width of the xo-canvas was also increased to accommodate the
change in size of adding in the new button to prevent overlap.


MORich-1:
Images of the execution:
Initial Click:
pr1
When OK is clicked:
pr2

MORich-1:
:+1:

deleteTempCache.php
```diff
@@ -0,0 +1,11 @@
+// This scipt is called by an administrator of the Formulize install within the Admin section of the site, in order to remove all files in the 'templates_c' folder.
```
eikonos: Comments need to be after the <?php tag because otherwise they'll be sent to the browser, as any text would.

deleteTempCache.php
```diff
@@ -0,0 +1,11 @@
...
+// This script should only be called if the user is developing new templates and needs a a quick way to delete existing cached template data.
+<?php
+    $files = glob($_SERVER["DOCUMENT_ROOT"].'/formulize/templates_c/*'); // Get all file names in directory.
```
eikonos: Rather than using DOCUMENT_ROOT, the XOOPS_ROOT_PATH variable should be used. The DOCUMENT_ROOT can point to different folders depending on the website configuration, so it's not really useful for finding the root folder of the Formulize site.

modules/formulize/admin/home.php
```diff
@@ -106,7 +106,8 @@ function readApplicationData($aid, $apps) {
 <a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=application&aid='.$aid.'&tab=screens"><i class="icon-screen"></i> Screens</a>
 <a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=application&aid='.$aid.'&tab=relationships"><i class="icon-connection"></i> Relationships</a>
 <a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=application&aid='.$aid.'&tab=menu%20entries"><i class="icon-menu"></i> Menu Entries</a>
-<a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=export&aid='.$aid.'"><i class="icon-download"></i> Export (beta!)</a>';
+<a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=export&aid='.$aid.'"><i class="icon-download"></i> Export (beta!)</a>
```
eikonos: This export button is contained inside each application, and the 'forms with no application', but it only needs to be shown once on the page, so it should probably appear next to the Formulize Preferences link.

modules/formulize/templates/admin/ui.html
```diff
@@ -160,6 +160,22 @@
     }
   });
 
+  $(".DCTemplates").click(function(){
+      var conf = confirm("Are you sure you want to Delete the template files? This should only be done when undergoing template development.");
+      if (conf) {
+          $.ajax({
+              url:'/formulize/deleteTempCache.php',
```
eikonos: Rather than linking directly to the file, the XOOPS_URL variable should be used because the path can change on different websites.

deleteTempCache.php
eikonos: Rather than adding this code in the deleteTempCache.php file, this code could be moved to the formulize_xhr_responder.php file as it deals with AJAX requests, loading the mainfile, checking permissions and so on.

MORich-1:
The export button has been moved, xhr_responder is now handling the new AJAX request, and document locations have been updated to use the xoops constants. Let me know if there is anything else to implement!
Here is a photo of the new placement:

newpos1

MORich-1:
:+1:

modules/formulize/formulize_xhr_responder.php
```diff
@@ -250,4 +251,15 @@
         }
     }
     break;
+
+    case "delete_template_cache":
+        // This scipt is called by an administrator of the Formulize install within the Admin section of the site, in order to remove all files in the 'templates_c' folder.
+        // This script should only be called if the user is developing new templates and needs a a quick way to delete existing cached template data.
+        $files = glob(XOOPS_ROOT_PATH.'/templates_c/*'); // Get all file names in directory.
```
eikonos: Should delete *.php files, so that the index.html file remains.

modules/formulize/templates/admin/ui.html
```diff
@@ -160,6 +160,22 @@
     }
   });
 
+  $(".DCTemplates").click(function(){
```
eikonos: It's not a big deal but "DCTemplates" isn't a great name because I don't know what it means. Something like "delete-templates" would be better.

modules/formulize/templates/admin/home.html
```diff
@@ -2,6 +2,7 @@
 <div style="float: right;">
 	<p><a href="../../system/admin.php?fct=preferences&op=showmod&mod=<{$adminPage.formulizeModId}>"><img src="../images/kedit.png"> <{$smarty.const._AM_HOME_PREF}></a></p>
 	<p style="text-align: right;"><a  href="ui.php?page=import">Import Application (beta!)</a></p>
+    <p style="text-align: right;"></a> <a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=export&aid='.$aid.'"><i class="icon-download"></i> Export (beta!)</a></p>
```
eikonos: Sorry, I see that I made a previous comment on the line for the Export link. The export link is per-application, so it needs to be inside each application block. The link/button to clear templates is per-site, not per-application, so it should be moved here instead.

modules/formulize/admin/home.php
```diff
@@ -106,7 +106,7 @@ function readApplicationData($aid, $apps) {
 <a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=application&aid='.$aid.'&tab=screens"><i class="icon-screen"></i> Screens</a>
 <a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=application&aid='.$aid.'&tab=relationships"><i class="icon-connection"></i> Relationships</a>
 <a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=application&aid='.$aid.'&tab=menu%20entries"><i class="icon-menu"></i> Menu Entries</a>
-<a href="'.XOOPS_URL.'/modules/formulize/admin/ui.php?page=export&aid='.$aid.'"><i class="icon-download"></i> Export (beta!)</a>';
+<button type = button class="DCTemplates">Delete Cached Templates</button>';
```
eikonos: Very minor detail, but for consistency with html formatting and the rest of the code this should be written as: type="button"

MORich-1:
:+1:

MORich-1:
Missed that one. Sorry about that! :+1:

PR #201, opened 2015-03-17
Revert from masquerade button

jnip:
The Issue, when masquerading is on:
 - (For me, on Chrome) clicking on the revert button did nothing.
 - The usual footer would not be exactly in the middle, it got shifted to the left

This fix adds a simple line break after the usual footer before displaying the revert button (if masquerading is on).

### DIFF
--- a/themes/formulize_standalone/theme.html
+++ b/themes/formulize_standalone/theme.html
@@ -134,6 +134,7 @@
     <br /> *}>
     <{$icms_footer}>
     <{ if $masquerade_username }>
+		<br>
         <div id="formulize-masquerade-block">
             Masquerading as <b><{$masquerade_username}></b>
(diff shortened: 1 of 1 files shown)


jnip:
:+1:

jnip:
@eikonos

themes/formulize_standalone/theme.html
```diff
@@ -134,6 +134,7 @@
     <br /> *}>
     <{$icms_footer}>
     <{ if $masquerade_username }>
+        <br>
```
eikonos: This puts the Revert button on a separate line, which I don't want, especially because then it appears by default just below the bottom of the page. I don't mind that the other footer text isn't in the absolute middle. The revert button works in Chrome and Firefox, but it doesn't seem to work in Safari. You may have to clear templates and force-refresh to get new CSS rules.

PR #203, opened 2015-03-20
radio_button-replace_blanks_with_a_default_value

jnip:
This should work now

### DIFF
--- a/modules/formulize/admin/save/element_options_save.php
+++ b/modules/formulize/admin/save/element_options_save.php
@@ -231,31 +231,6 @@ if(file_exists(XOOPS_ROOT_PATH."/modules/formulize/class/".$ele_type."Element.ph
   if($changed) {
     $_POST['reload_option_page'] = true; // force a reload, since the developer probably changed something the user did in the form, so we should reload to show the effect of this change
   }
-}
-
-
-// replace all existing blank radio button selections to the currently selected default
-if(isset($_POST['replaceblankswithdefault']) AND $_POST['replaceblankswithdefault']==1) {
-  if($ele_type == "radio" AND $default_entry = array_search(1, $element->getVar('ele_value'))) {
-    $sql = "SELECT `entry_id`, `".$element->getVar('ele_handle')."` FROM ".$xoopsDB->prefix("formulize_".$formObject->getVar('form_handle'));
-    if(!$res = $xoopsDB->query($sql)) {
-      print "Error replacing blank values for the options in element $ele_id";
-    } else {
-      $update_array = array();
-      // find the blank entries
-      while($res2 = $xoopsDB->fetchRow($res)) {
-	list($entry_id, $entry_value) = $res2;
-	if ($entry_value == NULL) {
-	  array_push($update_array, $entry_id);
-	}
-      }
-      // replace with default value
-      foreach ($update_array as $entry_id) {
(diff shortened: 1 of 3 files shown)


jnip:
@eikonos 
Could you take a look at the description I added (_AM_ELE_OPT_REPLACEBLANKSWITHDEFAULT).
Let me know if I need to change it. I will also need the description in french, to add in modules/formulize/language/french/admin.php

modules/formulize/admin/save/element_options_save.php
```diff
@@ -251,6 +251,31 @@
...
+      // find the blank entries
+      while($res2 = $xoopsDB->fetchRow($res)) {
+	list($entry_id, $entry_value) = $res2;
+	if ($entry_value == NULL || $entry_value == "") {
+	  array_push($update_array, $entry_id);
+	}
+      }
+      // replace with default value
+      foreach ($update_array as $entry_id) {
+	$sql = "UPDATE `".$xoopsDB->prefix("formulize_".$formObject->getVar('form_handle')."` SET `".$element->getVar('ele_handle')."`='".$default_entry."' WHERE entry_id=".$entry_id);
```
eikonos: Instead of using $default_entry directly in an SQL query, it should be  formulize_db_escape($default_entry) instead, because user input might contain text that's not safe for SQL.

modules/formulize/admin/save/element_options_save.php
```diff
@@ -251,6 +251,31 @@
...
+if(isset($_POST['replaceblankswithdefault']) AND $_POST['replaceblankswithdefault']==1 AND $ele_type == "radio") {
+  $default_entry = (isset($_POST['changeuservalues']) AND $_POST['changeuservalues']==1) ? array_search(1, $newValues) : array_search(1, $element->getVar('ele_value'));
+  if($default_entry) {
+    $sql = "SELECT `entry_id`, `".$element->getVar('ele_handle')."` FROM ".$xoopsDB->prefix("formulize_".$formObject->getVar('form_handle'));
+    if(!$res = $xoopsDB->query($sql)) {
+      print "Error replacing blank values for the options in element $ele_id";
+    } else {
+      $update_array = array();
+      // find the blank entries
+      while($res2 = $xoopsDB->fetchRow($res)) {
```
eikonos: Instead of looping through all the records, the rows could be updated with an SQL query like this:
UPDATE table SET field = 'New Value' where field = "" or field is NULL

modules/formulize/admin/save/element_options_save.php
```diff
@@ -251,6 +251,31 @@
   }
 }
 
+// replace all existing blank radio button selections to the currently selected default
+if(isset($_POST['replaceblankswithdefault']) AND $_POST['replaceblankswithdefault']==1 AND $ele_type == "radio") {
+  $default_entry = (isset($_POST['changeuservalues']) AND $_POST['changeuservalues']==1) ? array_search(1, $newValues) : array_search(1, $element->getVar('ele_value'));
```
eikonos: This code runs when the page is saved, and the new default value has not been stored in the database yet, so it's using the previously saved value from the database, which might be wrong.

jnip:
@eikonos I have applied the 3 mentioned changes. Please take another look.

jegelstaff:
A very useful feature that should be merged in. Conflicts should be simple to deal with.

jegelstaff:
See also: https://github.com/jegelstaff/formulize/pull/139

jegelstaff:
Addressed in #831

PR #202, opened 2015-03-19
Initial refactoring for cloning form screens: renaming of variable to al...

susannahkirby:
...low reuse/avoid duplication.

PR #204, opened 2015-03-22
Clone multipage screens

susannahkirby:
This builds on the work I did for cloning form screens.

I'm planning to do another round of refactoring since I may be able to pull much or all of this work into the parent screen.php class, which would make it all a lot more streamlined.


susannahkirby:
This should be good to go. It also contains all the commits from pull request #202 .

:+1:

susannahkirby:
:+1:

PR #206, opened 2015-03-25
Update handle when creating new form & element

jujinfu:
When creating new form or new element, if handle is left blank. It will use form/ element title replacing spaces with _. 
For element, handle will be Form_Handle_Element_Name for example.

### DIFF
--- a/modules/formulize/templates/admin/element_names.html
+++ b/modules/formulize/templates/admin/element_names.html
@@ -90,7 +90,7 @@ $(".savebutton").click(function() {
 			//kept this incase they didnt enter a new title, we can still update title when saving
 			var str="<{$content.formhandle}>"+"_"+$("[name=elements-ele_caption]").val();
 			str=str.toLowerCase().replace(new RegExp("[^a-z0-9]","gm"),"_");
-			str=str.replace(new RegExp("_{2,}","gm"),"_").substring(0,20);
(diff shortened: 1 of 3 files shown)


jujinfu:
:+1:

modules/formulize/templates/admin/element_names.html
```diff
@@ -80,12 +80,16 @@
...
 $(".savebutton").click(function() {
-  if($("[name=elements-ele_caption]").val() == "") {
+	if($("[name=elements-ele_caption]").val() == "") {
 		alert("Elements must have a caption!");
 		$("[name=elements-ele_caption]").focus();
 	}
 	<{* reload after ele_handle is blank, since it will be reset to the ele_id and that's a state change we'll need to reload to show the user *}>
 	if($("[name=elements-ele_handle]").val() == "") {
+		if ("<{$content.ele_id}>" == "new") {
+			var str="<{$content.formhandle}>"+"_"+$("[name=elements-ele_caption]").val().replace(new RegExp(" ","gm"),"_");
```
eikonos: This does replace spaces with _, but it should also make the text all lower-case and remove any punctation, basically, remove anything not A-Z or a-z or space.

jujinfu:
Now it will take the title and change it to all lowercase then replace any thing but a-z and 0-9 by _ . Next it will remove more than two _ eg. name____y will be name_y.  Finally it will take first 20 chars to use it as handle.

PR #208, opened 2015-03-25
formindex.php updated to create new formulize_screen_template table

susannahkirby:
@eikonos :
This isn't creating the new table in my db when I reload. Is there an update I need to run in my Modules Admin to get the new table patch in?


eikonos:
@susannahkirby If you go to this URL, it will probably create the table: http://www.yoursite.com/modules/formulize/admin/formindex.php?op=patch40

susannahkirby:
Hm, that's still not working. It says "DB updates completed. result: OK" but the table still isn't in the db afterwards.

modules/formulize/admin/formindex.php
```diff
@@ -916,6 +916,17 @@ function patch31() {
   INDEX i_sid (`sid`)
 ) ENGINE=MyISAM;";
                 }
+
```
eikonos: This block executes to upgrade from a much older version of Formulize. If you move this to around line 284, then it will run.

susannahkirby: Aha! Works now! Thank you!

susannahkirby:
This templateScreen.php class isn't complete, but I would welcome your feedback! 

In particular, I'm not 100% clear on how rendering works with screens, so I don't know if the render() implementation is headed in the right direction, especially with the code and template being pulled from a file. (In the meantime I'll also see if I can figure out how debugging works in PhpStorm.)

One other question at this point:

You said that "the code should be cached on disk, much like with the cache_on_before_save_code() function" but that "The code.php and template.html cache files should be stored in /modules/formulize/templates/screens/default/". To clarify - does this mean that they should NOT end up in the /formulize/cache folder (which is where the cache_on_before_save_code() goes)?

modules/formulize/class/templateScreen.php
```diff
@@ -0,0 +1,201 @@
...
+                return $screen;
+            }
+        }
+        return false;
+
+    }
+
+
+    function render($screen) {
+        include XOOPS_ROOT_PATH.'/header.php';
```
eikonos: The render function shouldn't include the header or the footer as those are included by the code which calls the render function.

eikonos:
The render function is correct except for the one note I made. Yes, the cached files should be stored in the location you've set now. They're kind of cached files, but all the other screen template cache files are stored in this location too, so it's more consistent.

susannahkirby:
@eikonos, this may be good to go. I had two questions outstanding:

1) Is there anything else in formindex.php that should be updated to reflect this new screen type? I don't know the previous versions of Formulize well enough to conceptualize how all the patches work.

2) Right now code.php and template.html are being updated (in the DB and in the cache) to reflect what the user has saved in the Templates tab. There's some boilerplate in there that seems like it should catch php errors, but it's not currently doing so. With that said, the same validation code (screen_list_templates.html) is also not catching errors in a listOfEntriesScreen Templates tab, and there is no code-checking being done in screen_multipage_templates.html. What's the best course of action, or should I leave this as-is?

modules/formulize/class/templateScreen.php
```diff
@@ -0,0 +1,259 @@
...
+
+
+    function render($screen) {
+        global $xoTheme;
+        if($xoTheme) {
+            $xoTheme->addStylesheet(self::FORMULIZE_CSS_FILE);
+            $xoTheme->addScript(self::FORMULIZE_JS_FILE);
+        }
+
+        $custom_code_filename = custom_code_filename($screen);
```
eikonos: The custom_code_filename(), template_filename(), and run_template_php_code() functions are all members of the class, so they should be called like $this->function().

susannahkirby:
Oh, and one last todo:

The constants
_AM_FORMULIZE_SCREEN_TEMPLATE_DESC_TEMPLATE
_AM_FORMULIZE_SCREEN_TEMPLATE_DESC_CUSTOM_CODE
should be updated to contain the actual text you want to display on a template screen's Templates tab.

Right now there's some filler text as the definition for each of these, since I didn't know what you wanted to say there.

eikonos:
This is really good! It works well, and the code is well organized and easy to read.

Issue #210, opened 2015-03-29
Uploading a file in a multipage form doesn't work


flaggada:
Hello

With last commits, it's not possible to upload a file, in a multipage form. You go on the next page but the file isn't uploaded.
Curiously, in a regular form (single page), no problem, you can do that.

Thanks

eikonos:
Pull request #211 fixes the file upload bug, and it was just merged into the master branch. :)

flaggada:
Cool thanks :-)

PR #211, opened 2015-03-31
Fix a miss matched element type

jujinfu:
Due to this miss matched type, when the file upload box is rendered, it
will be treated as xoopsFormLabel rather than XoopsFormFile object. When
it comes to xoopsFormLabel object, it will not contain
onchange="javascript:formulizechanged=1"  for file upload box, but this
part is the key to tell if there is a value changed.

I think this will fix the problem.

### DIFF
--- a/modules/formulize/class/elementrenderer.php
+++ b/modules/formulize/class/elementrenderer.php
@@ -1042,7 +1042,7 @@ class formulizeElementRenderer{
 			break;
 
 
-			case 'upload':
+			case 'fileUpload':
 				$form_ele = new XoopsFormFile (
 					$ele_caption,
 					$form_ele_id,
--- a/modules/formulize/class/elementrenderer.php
+++ b/modules/formulize/class/elementrenderer.php
@@ -1042,7 +1042,7 @@ class formulizeElementRenderer{
 			break;
 
 
-			case 'fileUpload':
+			case 'upload':
 				$form_ele = new XoopsFormFile (
 					$ele_caption,
 					$form_ele_id,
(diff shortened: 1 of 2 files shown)


jujinfu:
#210

modules/formulize/class/elementrenderer.php
```diff
@@ -1042,7 +1042,7 @@ function constructElement($form_ele_id, $ele_value, $entry, $isDisabled=false, $
 			break;
 
 
-			case 'upload':
+			case 'fileUpload':
```
eikonos: This doesn't fix the problem. The fileUpload type isn't handled by this code here, it's displayed by code in the fileUploadElement.php file.

jujinfu: I was checking the multi page one. The file upload didn't get saved because when the value of < intput type='file'...> changed, it didn't set the flag to 1 so that when we change page, the page will not know there is something changed, therefore, it's not saved. 

everything else got saved because they have a  condiction which is onchange="javascript:formulizechanged=1;"

in the javascript code, when we changing the page, it will check this var, if true then save.
That explains when i was having two elements when i change both, file gets uploaded. 

But you made a good point, i don't know where is this fileuploadelement.php get called, but it's better we add something like onchange="javascript:formulizechanged=1;" too.

jujinfu: I checked the file upload element object type is XoopsFormFile extends icms_form_elements_file, not formulizeFileUploadElement.

jujinfu:
:+1:

PR #212, opened 2015-03-31
Improve data exporting

eikonos:
Improve exporting by looping through batches of entries so that PHP time and memory limits are not exceeded. Also, now using the built-in PHP CSV output function.


eikonos:
:+1:

eikonos:
This has been tested and approved on a client site.

PR #215, opened 2015-05-25
Adding special cleaning function for ops

jegelstaff:
See note on the first (and only?) commit to know what the story is. This arose in CFCC.


jegelstaff:
:+1:

PR #218, opened 2015-09-18
ucosp changes pull request

jegelstaff:


modules/formulize/admin/ui.php
```diff
@@ -1,4 +1,7 @@
 <?php
+
+print "Hello World";
```
jegelstaff: Don't think we want to say hello here.

jegelstaff:
:+1:

jegelstaff:
Demo complete, closing now.

PR #225, opened 2015-09-19
parsed old commits, manually removed custom_code

JamesEarle:


jegelstaff:
We're actually leaving this in place, with some modifications in another branch.

PR #236, opened 2016-01-15
Synchronization Feature

jegelstaff:
Now: 

TODO: get it to work online??? (It has failed with large exports/short timeouts)
TODO: test inclusion of files!!! (Right now, it doesn't copy template files into position, but does have them in the .zip)
TODO: handle deletions!!! (deleted tables and elements don't happen)
TODO: is it possible to handle datatypes?? Export full create table syntax?? (datatyping is done in MySQL only, no metadata in Formulize for it)


modules/formulize/include/synchronization.php
```diff
@@ -0,0 +1,164 @@
...
+            return $archivePath;
+        }
+
+
+        include_once "../class/PDO_Conn.php";
+
+        /*
+         * syncTablesList function returns a complete list of database tables that are required to be synced
+         */
+        function syncTablesList() {
```
jegelstaff: We need to add a few other tables:

groups
group_permission
xoopsnotifications

modules/formulize/include/synchronization.php
```diff
@@ -0,0 +1,226 @@
...
+         *          
+         *      1. add folder stucture to template file part of archive
+         */
+        
+        /*
+         * doExport function exports template files and current Formulize database state to a ".zip" archive
+         * 
+         * param archiveName        string representing path to new or existing zip file. path must have ".zip" extension
+         */
+        function doExport($archiveName){
```
jegelstaff: Makes sense to have doExport receive a preferred filename, also might make sense for this to have an optional parameter for the folder, but default to the Export folder if nothing specified.

For that matter, might make sense for the filename to also be optional, and default inside the function to something sensible if no filename is passed in.

PR #240, opened 2016-03-28
Name cloned forms

jegelstaff:


modules/formulize/templates/admin/application_forms.html
```diff
@@ -30,6 +33,8 @@
 });
 
 $(".cloneform").click(function () {
+	var formName = prompt("Enter a name for the cloned form:", "");
```
jegelstaff: We need to put all this in an IF, so if they don't provide a name, then we do nothing.

jegelstaff:
Possibly superseded by 275

PR #253, opened 2017-09-22
Display correct success or failure icon beside installation requirements

lelbarton:
The image yes.png was displayed for all requirements regardless of whether they had been met or not. This implements conditionals to display no.png if there is a problem.

### DIFF
--- a/install/page_modcheck.php
+++ b/install/page_modcheck.php
@@ -79,7 +79,7 @@ function successImg($success_val) {
 }
 
 $php_version_success = phpVersionCheck();
-$mysql_success = function_exists('mysql_connect') ? 1 : -1;
+$mysql_success = function_exists('mysqli_connect') ? 1 : -1;
 $extension_success = extension_loaded('session') ? 1 : -1;
 $pcre_success = extension_loaded('pcre') ? 1 : -1;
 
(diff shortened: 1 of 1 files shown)


install/page_modcheck.php
```diff
@@ -26,135 +26,151 @@
...
-<fieldset>
-<h3><?php echo RECOMMENDED_EXTENSIONS; ?></h3>
-<p><?php echo RECOMMENDED_EXTENSIONS_MSG; ?></p>
-<div class="clear">&nbsp;</div>
+    <fieldset>
+        <h3><?php echo REQUIREMENTS; ?></h3>
+        <h4><?php echo SERVER_API; ?>:&nbsp; <?php echo php_sapi_name(); ?> <img
+                    src="img/yes.png" alt="Success" class="rootimg" /></h4>
+        <div class="clear">&nbsp;</div>
+        <h4><?php echo _PHP_VERSION; ?>:&nbsp;
```
jegelstaff: Isn't the constant PHP_VERSION? no underscore before it?

lelbarton: This was what was in the file originally — maybe a previously committed typo?

PR #254, opened 2017-09-22
Fix for crash when using last modified user search

Cirieneve:


jegelstaff:
Looks good.

So, what would make it better, is if we were not relying on hard coding all the metadata field names into this file!

We can get around this conveniently by using a property of the data class in Formulize. This code gives us an array of the metadata fields:

    $dataHandler = new formulizeDataHandler();
    $metadataFields = $dataHandler->metadataFields;

A couple wrinkles though...those names are all defined in all caps in the file, which is kind of silly. They are only ever referenced as lower case names elsewhere in the code. The data handler class is in the modules/formulize/class/data.php file.

As part of that case change, you should search the code for other references to metadataFields and fix the converting of things to upper case and lower case that happens right now, because the values returned are currently all upper case.

So, next issue is that we need to have information about that the data type of each metadata field is, text or date. I suggest we change the structure of this somehow, so the types are available/included in the array that gets returned.

We could just change it so the metadata field names are the keys of the array and the values of each are the data types. But that is somewhat inelegant to use, because you cannot then simply do a foreach on the array and get what you want. You have to foreach the keys of the array.

The most correct approach may be to use the iterator structures in PHP, and make this a custom object and iterate over it in a special way.... but that smacks of overengineering to me even though it is "correct".

We can talk about the right approach and figure it out.

jegelstaff:
foreach(keys($metadataFields) as $name) {
  print "datatype is: ".$metadataFields[$name];
}

jegelstaff:
So if we convert it to key value pairs, then you iterate over the keys, and use the key to access the value in the array.

But, from a semantic readability point of view, that's a mess.

Perhaps we should really relabel everything too....

$metadataFieldTypes = array('creation_uid'=>'text', etc...);

foreach(keys($metadataFieldTypes) as $field) {
  print "datatype is: ".$metadataFieldTypes[$field];
}

jegelstaff:
Frankly.... perhaps we should add another array.... we want this metadataFieldTypes array in the data class to support what you're doing with this fix. But we would probably do better with the metadataFields array that was simply the field names, in the other places where it is currently used, since they don't need the type information at all.

jegelstaff:
In both arrays, everything would be lower case.

jegelstaff:
Superseded by pull request 256

PR #255, opened 2017-09-23
Convert mysql function calls to mysqli equivalents in install process

lelbarton:


install/class/dbmanager.php
```diff
@@ -20,6 +20,9 @@
  * @version $Id: dbmanager.php 20502 2010-12-08 00:27:17Z skenow $
  * @access public
  **/
+
+include_once '..\libraries\icms\db\legacy\PdoDatabase.php';
```
jegelstaff: Don't need this one anymore.

install/page_dbconnection.php
```diff
@@ -46,8 +46,8 @@
 
 $error = '';
 if ($_SERVER['REQUEST_METHOD'] == 'POST' && !empty( $vars['DB_HOST'] ) && !empty( $vars['DB_USER'] )) {
-	$func_connect = empty( $vars['DB_PCONNECT'] ) ? "mysql_connect" : "mysql_pconnect";
-	if (! ( $link = @$func_connect( $vars['DB_HOST'], $vars['DB_USER'], $vars['DB_PASS'], true ) )) {
+    $hostname = empty( $vars['DB_PCONNECT'] ) ? $vars['DB_HOST'] : "p:".$vars['DB_HOST'];
+    if (! ( $link = @mysqli_connect( $hostname, $vars['DB_USER'], $vars['DB_PASS'], true ) )) {
```
jegelstaff: DB name needs to be added here, fourth param is the db name, I think the 'true' is leftover from the old syntax...probably don't need that at all? See what the other params are and if anything further past db name makes sense to include.

install/page_dbsettings.php
```diff
@@ -28,7 +28,7 @@
 
 $vars = & $_SESSION ['settings'];
 
-$func_connect = empty ( $vars ['DB_PCONNECT'] ) ? "mysql_connect" : "mysql_pconnect";
+$func_connect = empty ( $vars ['DB_PCONNECT'] ) ? "mysqli_connect" : "mysqli_pconnect";
```
jegelstaff: This needs to be changed as in the other file, once everything is working (I'm assuming the connection page is before the settings page).

install/page_dbsettings.php
```diff
@@ -28,8 +28,8 @@
 
 $vars = & $_SESSION ['settings'];
 
-$func_connect = empty ( $vars ['DB_PCONNECT'] ) ? "mysql_connect" : "mysql_pconnect";
-if (! ($link = @$func_connect ( $vars ['DB_HOST'], $vars ['DB_USER'], $vars ['DB_PASS'], true ))) {
+$hostname = empty ( $vars ['DB_PCONNECT'] ) ? $vars ['DB_HOST'] : "p:".$vars ['DB_HOST'];
+if (! ($link = @mysqli_connect( $hostname, $vars ['DB_USER'], $vars ['DB_PASS'], $vars ['DB_NAME'] ))) {
```
jegelstaff: I think you can leave the name out here, and rely on the select db part below

install/page_dbsettings.php
```diff
@@ -167,9 +167,10 @@ function xoFormBlockCollation($name, $value, $label, $help = '', $link, $charset
 	$error = validateDbCharset ( $link, $vars ['DB_CHARSET'], $vars ['DB_COLLATION'] );
 	$db_exist = false;
 	if (empty ( $error )) {
-		if (! @mysql_select_db ( $vars ['DB_NAME'], $link )) {
+		if (! @mysqli_select_db ($link, $vars ['DB_NAME'] )) { // TODO update
```
jegelstaff: This will set the active db, or attempt to create a new db with that name if none exist, so if we specifically connect to a db by name above, this logic will be busted I think. Hence removing db name in the connection line.

install/page_dbconnection.php
```diff
@@ -46,8 +46,8 @@
 
 $error = '';
 if ($_SERVER['REQUEST_METHOD'] == 'POST' && !empty( $vars['DB_HOST'] ) && !empty( $vars['DB_USER'] )) {
-	$func_connect = empty( $vars['DB_PCONNECT'] ) ? "mysql_connect" : "mysql_pconnect";
-	if (! ( $link = @$func_connect( $vars['DB_HOST'], $vars['DB_USER'], $vars['DB_PASS'], true ) )) {
+    $hostname = empty( $vars['DB_PCONNECT'] ) ? $vars['DB_HOST'] : "p:".$vars['DB_HOST'];
+    if (! ( $link = @mysqli_connect( $hostname, $vars['DB_USER'], $vars['DB_PASS']) )) {
 		$error = ERR_NO_DBCONNECTION;
```
jegelstaff: You should be able to put in print mysqli_error() at this point, or some function like that, which will tell us the error from MySQL, such as invalid host or something. What you're seeing on screen now I think is the installer itself saying there's no connection, but we need to see why.

jegelstaff:
Tested on PHP 5, works. :)

PR #256, opened 2017-09-23
Fix hardcoded metadata fields

Cirieneve:


jegelstaff:
mainfile.php should not be part of the branch's changes. It will be different in your installation compared to the master repo, but that's because it is unique for every installation of Formulize. It is built at the time you install.

modules/formulize/class/data.php
```diff
@@ -39,6 +39,7 @@ class formulizeDataHandler  {
 	
 	var $fid; // the form this Data Handler object is attached to
 	var $metadataFields; //
+    var $metadataFieldType;
```
jegelstaff: Let's make it plural

modules/formulize/class/data.php
```diff
@@ -49,14 +50,22 @@ function formulizeDataHandler($fid){
...
 		
-		//set the avaiable metadata fields to a global
-		$this->metadataFields = array("ENTRY_ID", 
-		  						"CREATION_DATETIME",
-		   						"CREATION_UID",
-		   						"CREATOR_EMAIL",
-		  						"MOD_DATETIME",
-		   						"MOD_UID",
-		   						"USER_VIEWEMAIL");  
+		//set the available metadata fields to a global
```
jegelstaff: We can get rid of user_viewemail, I think that was put in here by mistake originally.

modules/formulize/include/entriesdisplay.php
```diff
@@ -4575,11 +4575,13 @@ function formulize_gatherDataSet($settings=array(), $searches, $sort="", $order=
             // used for trapping the {BLANK} keywords into their own space so they don't interfere with each other, or other filters
             $addToItsOwnORFilter = false;
 
-            if ("creation_uid" == $key OR "entry_id" == $key) {
-                $ele_type = "text";
-            } else {
-                $elementObject = $element_handler->get($key);
-                $ele_type = $elementObject->getVar('ele_type');
+            $dataHandler = new formulizeDataHandler(false);
```
jegelstaff: OK, so this needs work, rather than printing out the data, we need to mimic the original behaviour that fixed the error as you had originally. Just we pull the names from the array now and iterate over them, rather than having them hard coded in the if statement. This means that you need to check for the presence of the $key in the array. You can use isset() for that, as in: if(isset($metadataFieldTypes[$key]))....

modules/formulize/include/functions.php
```diff
@@ -5611,12 +5611,11 @@ function formulize_findCommonValue($form1, $form2, $key1, $key2) {
 
 //Function used to check if the given field is within the list of metadata fields.
 function isMetaDataField($field){
-    $ucField = strtoupper($field);
     $dataHandler = new formulizeDataHandler(false);
     $metadataFields = $dataHandler->metadataFields;
     foreach ($metadataFields as $value) 
     {
-        if($value == $ucField)
+        if($value == $field)
```
jegelstaff: Good, and I think there was one other place that this array was being referenced that can have a tiny refactor like this, just search the codebase for metadataFields.

modules/formulize/include/entriesdisplay.php
```diff
@@ -4578,10 +4577,14 @@ function formulize_gatherDataSet($settings=array(), $searches, $sort="", $order=
             $dataHandler = new formulizeDataHandler(false);
             $metadataFieldTypes = $dataHandler->metaDataFieldTypes;
 
-            foreach (array_keys($metadataFieldTypes) as $field) {
-                print $field;
-                print $metadataFieldTypes[$field];
-                $ele_type = $metadataFieldTypes[$field];
+            if (isset($metadataFieldTypes[$key])){
```
jegelstaff: Well, you don't want to look through the entire array... look at the code there, you will end up with $ele_type equal to the type of the last field in the array. You want to simply set $ele_type to the type of the $key... $ele_type = $metadataFieldTypes[$key];

modules/formulize/include/entriesdisplay.php
```diff
@@ -4575,9 +4574,13 @@ function formulize_gatherDataSet($settings=array(), $searches, $sort="", $order=
             // used for trapping the {BLANK} keywords into their own space so they don't interfere with each other, or other filters
             $addToItsOwnORFilter = false;
 
-            if ("creation_uid" == $key OR "entry_id" == $key) {
-                $ele_type = "text";
-            } else {
+            $dataHandler = new formulizeDataHandler(false);
+            $metadataFieldTypes = $dataHandler->metaDataFieldTypes;
```
jegelstaff: metaDataFieldTypes should be metadataFieldTypes to be consistent with the declaration in the class file

jegelstaff:
I think this looks good. Make sure you've tested searching on the modification user column to be sure.

Then read up about the test process, install Selenium Builder in Firefox, and we'll talk about making a test for this.

jegelstaff:
I still see .gitignore and mainfile.php in the changed files. Probably just add those two filenames to your .gitignore, and there you go.

ci/selenium2-tests/009-metadata-fields-search.json
```diff
@@ -0,0 +1,488 @@
+{
```
jegelstaff: Looks OK, but will need some test steps to check if the desired text/entries are on screen after the search and we haven't got a blank list. That can wait.

jegelstaff:
Also, see comment on the test file, for a later date. Not urgent.

PR #263, opened 2017-11-29
Handles formatting for empty templates - currently untested

Cirieneve:


jegelstaff:
Just needed the path added to the file reading and writing functions. Otherwise, works like a charm. :)

PR #268, opened 2018-01-27
super rough transition, can login with google but not to correct page…

knewbury01:
… with correct user state yet... todos: 
transition to correct page
get user email and connect to current login interface to id user by email
fix hardcoded redirect url and secret file path, 
need doc for howto acquire secret file/do setup for users on google account end of things...


jegelstaff:
Oh! By Thumbs up, I meant 👍

jegelstaff:
As in, a new comment, rather than reaction.

jegelstaff:
Because of the new library, the changes are too big to allow commenting on files directly. The UI here doesn't provide access to that many files. So some comments about particular files then...

session.php - where the $externalId, etc, is checked and so on... I am not certain that the IF conditions are sufficient to make that work the way we want in all cases. It's checking $_GET for the code and newcode, and it processes some data from $_SESSION or from Google depending on what it finds, and then it does other stuff. But there's no validation that it actually got what it expects out of $_SESSION, or got back anything real from Google. And then it goes straight into user creation if no resource mapping is found. I think this needs to be beefed up a bit security-wise. I think right now, if you simply put any value in the URL for ?code=xyz then it would trigger the creation of a user, and that's probably not what we want. Even if that's not the case, there may be other ways to circumvent this code because it doesn't check if $userData that it gets from Google is what we think it should be, or that there even is an e-mail in $_SESSION.

formindex.php - nice work adding in the database patching. However, you don't need to add in the specific checks for the new table. Simply declaring the table name as $checkThisTable (replacing apikeys) will do the trick. The big comment above that line explains the checking logic at work in the file. It's dynamic, based on the declared table name that represents the most recent change to the database.

Very cool that you added docs. :) :+1: 

For the token management UI, a few things I'd change: there can be hundreds of groups in some sites, so I would make that a multi-selection list box, rather than checkboxes, to save space. Or maybe switch to a multi-selection list box, if there are more than say 15 groups, because checkboxes are better for small numbers of choices. Also, I'd use a plain textbox for the number of uses, so any value can be entered there.

Overall, looks good! Thank you!

PR #269, opened 2018-01-31
Copying old UCOSP changes into Git

Jyjhuang:
add graph screen type for modules/formulize/admin/save/screen_settings_save.php file


generateUser.php
```diff
@@ -0,0 +1,23 @@
+<?php
```
jegelstaff: This file should be removed from the branch

include/readelements.php
```diff
@@ -0,0 +1,541 @@
+<?php
```
jegelstaff: This file belongs in the modules/formulize/include folder.

modules/formulize/admin/formindex.php
```diff
@@ -494,6 +507,78 @@ function patch40() {
             print "<p>If you have any questions about this upgrade issue, please contact <a href=mailto:formulize@freeformsolutions.ca>Freeform Solutions</a> for assistance.</p>";
         }
 
+        $screenpathname = XOOPS_ROOT_PATH."/modules/formulize/templates/screens/default/";
```
jegelstaff: This block already exists in formindex.php, with some changes. But I believe the current version should supersede the old one entirely, so this block can be removed.

modules/formulize/admin/formindex.php
```diff
@@ -643,6 +728,49 @@ function patch40() {
     }
 }
 
+function saveMenuEntryAndPermissionsSQL($formid,$appid,$i){
```
jegelstaff: This block also already exists in formindex.php and doesn't need to be added again.

modules/formulize/class/screen.php
```diff
@@ -47,6 +49,47 @@ function __construct() {
...
 
+  /**
+   * Checks if a given template name exists for the current screen.
+   * @return true if the given template name exists for the screen, false otherwise.
+   */
+  function hasTemplate($templateName) {
+      return (trim($this->getTemplate($templateName)) != "");
+  }
+
+  function getTemplate($templateName) {
```
jegelstaff: getTemplate already exists in this file. The current version should supersede this one, so this one can be removed.

modules/formulize/include/elementdisplay.php
```diff
@@ -220,11 +220,36 @@ function displayElement($formframe="", $ele, $entry="new", $noSave = false, $scr
             $isDisabled = !formulizePermHandler::user_can_edit_entry($form_id, $user_id, $entry);
 		}
 
+		/*
```
jegelstaff: This block is dealing with permissions, which in the existing file is now handled by line 220. The same comment introduces that condition and this one. This block can be removed.

modules/formulize/include/entriesdisplay.php
```diff
@@ -1151,109 +1153,14 @@ function drawInterface($settings, $fid, $frid, $groups, $mid, $gperm_handler, $l
 			print "</table></div>";
 		}
 
-		print "<table cellpadding=10><tr><td id='titleTable' style=\"vertical-align: top;\" width=100%>";
```
jegelstaff: There should be a new file corresponding to these deleted lines. These lines were moved to be the "default template." I didn't see a new file added as part of this branch, so that needs to be done. Also, whatever the actual default template is in the UCOSP branch, the content of that file should now be all these lines removed from the current github copy of entriesdisplay.php, because it is the current functionality that we want to use as the current default template behaviour.

class/data.php
```diff
@@ -0,0 +1,357 @@
+<?php
```
jegelstaff: I don't understand where this file is coming from or what it is used for. I don't recall that we ever worked on a core file like this, only Formulize files. In fact, the UCOSP branches don't include "core" files from outside modules/formulize. So what is this from?

include/readelements.php
jegelstaff: Actually, I see you have a copy of this with changes in the right place lower down. So this copy can just be removed from the branch.

modules/formulize/include/formdisplay.php
```diff
@@ -49,155 +49,178 @@
 
 // NEED TO USE OUR OWN VERSION OF THE CLASS, TO GET ELEMENT NAMES IN THE TR TAGS FOR EACH ROW
 class formulize_themeForm extends XoopsThemeForm {
-
```
jegelstaff: I'm not sure why apparently the same code is showing as a removal and then an addition in this file? Is that a whitespace issue? Check that these methods in side the formulize_themeForm class have not actually changed, or if they have, isolate the changes to just the lines that have changed.

modules/formulize/include/formdisplay.php
```diff
@@ -883,25 +928,75 @@ function displayForm($formframe, $entry="", $mainform="", $done_dest="", $button
 					$form->addElement($subLinkUI);
 				}
 			}
-		}
```
jegelstaff: Another case where it looks like the added code includes some of the removed code?

modules/formulize/include/formdisplay.php
```diff
@@ -1970,21 +2214,17 @@ function compileElements($fid, $form, $formulize_mgr, $prevEntry, $entry, $go_ba
...
 	$gridCounter = array();
 	$inGrid = 0;
-	
+
 	formulize_benchmark("Ready to loop elements.");
 
 	// set the array to be used as the structure of the loop, either the passed in elements in order, or the elements as gathered from the DB
 	// ignore passed in element order if there's a screen in effect, since we assume that official element order is authoritative when screens are involved
 	// API should still allow arbitrary ordering, so $element_allowed can still be set manually as part of a displayForm call, and the order will be respected then
-	if(!is_array($elements_allowed) OR $screen) {
```
jegelstaff: In this case, the existing code should supersede the UCOSP branch code.

modules/formulize/include/functions.php
```diff
@@ -4873,6 +4873,11 @@ function buildConditionsFilterSQL($conditions, $targetFormId, $curlyBracketEntry
                 } else {
                     $conditionsfilter .= " AND ";
                 }
+
```
jegelstaff: This block and the next one... where is $newFilterOps coming from? I think this code is nonfunctional, or meant to do something different from what the current code makes possible at this point.

jegelstaff: Does the _buildConditionsFilterSQL return different values in the UCOSP branch perhaps?

modules/formulize/include/functions.php
```diff
@@ -4995,12 +5004,17 @@ function _buildConditionsFilterSQL($filterId, &$filterOps, &$filterTerms, $filte
       } else {
 	$filterOps[$filterId] = 'IS';
       }
-    }
```
jegelstaff: Again, I'm not sure what the replacement code in this case is meant to do...this is a part of the code that has been perfected a fair bit over recent years, and I don't think we were doing anything with it in the functionality the UCOSP branch was developing. Since this doesn't relate to the template work, I would pull this and the newFilterOps changes above, out of the branch.

modules/formulize/include/functions.php
```diff
@@ -5320,6 +5334,130 @@ function generateHiddenElements($elements, $entry) {
...
+        $member_handler = & xoops_gethandler('member');
+    }
+    $form_meta = getMetaData($entryId, $member_handler, $fid);
+    $breakHTML = "<b>" . _formulize_FD_ABOUT . "</b><br>";
+    $breakHTML .= _formulize_FD_CREATED . $form_meta['created_by'] . " " . formulize_formatDateTime_duplicate_from_formdisplay_php($form_meta['created']) . "<br>" . _formulize_FD_MODIFIED . $form_meta['last_update_by'] . " " . formulize_formatDateTime_duplicate_from_formdisplay_php($form_meta['last_update']);
+    return $breakHTML;
+}
+
+// THIS FUNCTION FORMATS THE DATETIME INFO FOR DISPLAY CLEANLY AT THE TOP OF THE FORM
+// Old function is kept in formdisplay.php for historical reasons so nothing will break
```
jegelstaff: This comment makes it sound like we're purposely leaving duplicate code or functions around. We should just have one copy of the function. And it doesn't need a strange name like this. It's okay to move a function from formdisplay.php into functions.php. As long as the name is the same, the calling code will find the function. So verify that the two functions are the same, and once made equivalent, if any work is required to do so, then put the canonical copy in functions.php and make sure that everywhere that calls either function now calls the canonical copy in functions.php.

modules/formulize/include/functions.php
```diff
@@ -5375,18 +5513,18 @@ function formulize_db_escape($value) {
 }
 
 // THANKS TO baptiste.place@utopiaweb.fr on php.net for this conversion function:
-/**
```
jegelstaff: Another odd removal/addition block where the code looks the same?

generateUser.php
Jyjhuang: removed

include/readelements.php
Jyjhuang: removed

modules/formulize/admin/formindex.php
Jyjhuang: removed

modules/formulize/admin/formindex.php
Jyjhuang: removed

modules/formulize/class/screen.php
Jyjhuang: removed

modules/formulize/include/elementdisplay.php
Jyjhuang: removed the comment block

class/data.php
Jyjhuang: removed

modules/formulize/include/formdisplay.php
Jyjhuang: There are differences between the current iteration and the UCOSP branch for the formulize_themeForm class. I'm not sure why it only marked the line here though.

modules/formulize/include/formdisplay.php
Jyjhuang: same as above.

modules/formulize/include/formdisplay.php
Jyjhuang: reverted

modules/formulize/include/functions.php
Jyjhuang: Yup, there are some quite significant changes for _buildConditionsFilterSQL between the current one and UCOSP branch. We can talk about what to do here?

Jyjhuang: Yup, there are some quite significant changes for _buildConditionsFilterSQL between the current one and UCOSP branch. We can talk about what to do here?

Jyjhuang: pulled out the newFiltersOps for now

modules/formulize/include/functions.php
Jyjhuang: same as above

modules/formulize/include/functions.php
Jyjhuang: removed the duplicate method in formdisplay.php

modules/formulize/include/functions.php
Jyjhuang: I believe here it's just spacing issues.

modules/formulize/include/entriesdisplay.php
Jyjhuang: TBD

PR #274, opened 2018-09-23
Add a new Relationship  between forms using Primary key - Foreign key Relation, compared to currently using linked elements.

chubeyj3:
Add a constant String value to mainfile.php to describe the new textbox
selection. The new textbox will only be displayed if the first key is
not equal to an ID.

### DIFF
--- a/modules/formulize/class/forms.php
+++ b/modules/formulize/class/forms.php
@@ -425,7 +425,7 @@ class formulizeFormsHandler {
 		return new formulizeForm();
 	}
 
-	function &get($fid,$includeAllElements=false) {
+	function get($fid,$includeAllElements=false) {
 		$fid = intval($fid);
 		// this is cheap...we're caching form objects potentially twice because of a possible difference in whether we want all objects included or not.  This could be handled much better.  Maybe iterators could go over the object to return all elements, or all visible elements, or all kinds of other much more elegant stuff.
 		static $cachedForms = array();
--- a/modules/formulize/class/forms.php
+++ b/modules/formulize/class/forms.php
@@ -1402,21 +1402,37 @@ class formulizeFormsHandler {
 			$res = $this->db->queryF($sql);
 }
 	}
-        function hasForeignKeyLinkElement($fid){
+        
+        /**
+         * This function goes through the "formulize" table (which is really the
+         * elements table), and retrieves all elements in the form with form id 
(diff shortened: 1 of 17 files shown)


modules/formulize/class/forms.php
```diff
@@ -1402,4 +1402,22 @@ public function setPermissionsForClonedForm($fid, $newfid)
 			$res = $this->db->queryF($sql);
 }
 	}
+        function hasForeignKeyLinkElement($fid){
+            global $xoopsDB;
+            $sql = "SELECT ele_foreign_key_element_link FROM " . $xoopsDB->prefix("formulize") . " WHERE id_form=" . $fid . ";";
```
jegelstaff: The SQL can be made a bit better by adding ele_foreign_key_element_link = 1 to the WHERE clause, then just checking if the $xoopsDB->getRowsNum($res) > 0 after running the query.

jegelstaff:
I suggest renaming ele_foreign_key_element_link to ele_foreign_key_element. I don't think 'link' adds anything to the semantics. It's just a flag to say if this is a foreign_key_element or not.

jegelstaff:
I think what we therefore need, is multiple foreign key elements per form, one for each link to another form that this form has. To track this and identify if the link already exists or not, we should add the ID number of the other form to the handle name when creating the element:

foreign_key_from form_12
foreign_key_from_form_58

Etc.

This is NOT the name of the newly added field in the 'formulize' table. That is OK as is (and just remove _link from the end). This is the ele_handle of the elements that get created. We need to reference the form they tie back to, since there might be multiple connections.

Then, when checking if an element exists, it's not just whether a foreign key element exists on this form, it's whether one exists, that points back to the same other form.

Which further suggests that instead of a 1/0 flag, the ele_foreign_key_element property of the element, should contain the ID of the form it's pointing at. That is a nice authoritative way to easily keep track of that. And the id should also be appended to the end of the element's ele_handle so we can ensure they remain unique, and easily readible.

jegelstaff:
This was an interesting idea at the time. But pollutes the database with extra hidden stuff that is "magic." We should have the magic be driven by the interface, and create structures for the user, which they can interact with and adjust as desired if necessary. This is also probably superseded by various improvements to the generation of subform elements, and enforcement of one-to-one links.

Issue #172, opened 2015-02-02
Cloned forms don't allow renaming of form until post-creation.


Torboto:
Should ask for name of new form after clicking 'Clone' button, or something.

PR #275, opened 2018-09-23
Add prompt for new name when cloning forms

anisha-rohra:
- Fixed bug involving ability to view cloned forms on admin view
- Changed behaviour for viewing cloned forms on menu from not visible to visible by default

Issue referenced here: https://github.com/jegelstaff/formulize/issues/172


jegelstaff:
Need to check against 240

PR #342, opened 2023-01-11
More fixes for Fatal Error on Formulize admin side

tikyon:


libraries/icms/core/DataFilter.php
```diff
@@ -980,7 +980,7 @@ static private function priv_checkVar($data, $type, $options1, $options2) {
 					break;
 
 					case "host":
-						$valid = filter_var($data, FILTER_VALIDATE_URL, FILTER_FLAG_HOST_REQUIRED);
+						$valid = filter_var($data, FILTER_VALIDATE_URL);
```
jegelstaff, line 983: I'm curious about the situation here. It makes sense that if the validation operation is looking for 'host' that we would want the host to be required. But I assume you have an unconventional URL? Can you give an example of where this code is/was problematic?

jegelstaff:
Look like you're using the Pages feature in the Content module? That's cool. Steve Kenow @skenow in Github, is one of the key maintainers of ImpressCMS and would know more about that module if there are deeper questions. Unfortunately, the copy of it in Formulize is a forked version from ImpressCMS 1.3, because starting with a later version they did some really unhelpful things in the database layer that meant we basically could not take the code upstream, so had to fork. But on the plus side, the 1.3 code base was never upgraded to PHP 8, at least I don't think so. So what we've got is further advanced than the original in that sense.

libraries/icms/core/DataFilter.php
tikyon: https://php.watch/versions/8.0/filter_var-flags 
The FILTER_FLAG_HOST_REQUIRED flag got removed in PHP8, so it throws fatal error if we keep it.

jegelstaff: Aha, I wondered about that. Did not see a reference in the documentation. But then, there it is, in the notes at the bottom of the page. Cool.

PR #369, opened 2023-11-02
BUGFIX: Fixing issue where saving a form would null existing file uploads.

arvinsingla:
When editing a form that contains a subform with existing file uploads. The file would get null'd in the database when a save operation happens. This is a simple brute force fix that fetches the existing value from the DB and sends that to get written whenever a save happens preserving the original value.


jegelstaff:
An "abort writing" flag would be useful, but that is a larger architectural change. This is good.

PR #372, opened 2023-11-10
Fix/build filter improvements

jegelstaff:
Fixes construction of filters with buildFilter under certain conditions, includes documentation of the function.

### DIFF
--- a/docs/function-buildFilter.md
+++ b/docs/function-buildFilter.md
@@ -1,0 +1,90 @@
+---
+layout: default
+permalink: developers/API/functions/buildFilter/
+---
+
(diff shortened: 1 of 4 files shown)


jegelstaff:
Still needs some refinements around multi (checkbox) filters:

- [ ] no checkbox should be shown on top option
- [ ] default text for top option should be different
- [ ] checkboxes and labels need to align properly (CSS inside/outside issue? or positioning of label?)

jegelstaff:
Major cleanup and rationalization of the logic is possible. But it's working, and is a bit streamlined.

PR #374, opened 2023-11-10
FEATURE/Remote-Inclusion

jegelstaff:
remote.html goes into other website that you want to pull Formulize content into.

Much to do here. But possibly this could just be used as the way Formulize works in general? ie: have javascript calls handle all submissions and reloads, get away from the architecture that relies on the browser behaviour of submitting forms and reloading pages? Then there would be just one unified mechanism of action for all page display and submission handling.

### DIFF
--- a/modules/formulize/include/entriesdisplay.php
+++ b/modules/formulize/include/entriesdisplay.php
@@ -3336,7 +3336,7 @@ function addNew(flag) {
 	} else {
 		window.document.controls.ventry.value = 'addnew';
 	}
-	window.document.controls.submit();
+	submitControls();
 }
 
 function goDetails(viewentry, screen) {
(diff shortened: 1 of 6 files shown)


jegelstaff:
No where near ready for merge!

jegelstaff:
Todo:

- [ ] Merge remote CSS and nest in a selector for the id where the markup will go
- [ ] Does keycode 13 need to be trapped only on list screens not all screens?
- [ ] Make CORS locations configurable
- [ ] when remotesubmitform is potentially called, the alternate target needs to be formulize_mainform not controls!
- [ ] Get icons working
- [ ] Test list popups (change cols, save view, etc)
- [ ] Test fullform subforms
- [ ] Test modal subforms

PR #375, opened 2023-11-10
Further PHP 8 typing issues and other fixes from OCANDS

jegelstaff:
Changes pulled from the OCANDS repo. Want to make sure they're in 7.2 imminently.

Basically, just want to know if this looks complete, or was there more work that went into the code that we should include?


jegelstaff:
Basically, just want to know if this looks complete, or was there more work that went into the code that we should include?

tikyon (commented):
The other three files look good.

Added two more file changes for entriesdisplay.php and main.php

jegelstaff:
Thank you!

Issue #389, opened 2023-11-14
Printable view showing JavaScript code when using subform


tikyon:
When using subform of collapsible or flat form types, the script is displayed on the page before rendering the subform rows. It doesn't happen when displaying the subform in table rows.

The script shown is from file modules/formulize/include/formdisplay.php  [around Line 2567](https://github.com/jegelstaff/formulize/blob/16353fa8cc674daaa5936b9a259111f5cc05bf87/modules/formulize/include/formdisplay.php#L2567) and [Line 2590](https://github.com/jegelstaff/formulize/blob/16353fa8cc674daaa5936b9a259111f5cc05bf87/modules/formulize/include/formdisplay.php#L2590).

I suspect that the opening `<script>` was not included properly.

jegelstaff:
Ack! I guess 7.21 is closer to reality that I thought.

I have made a PR that I think fixes this. Please check it out: https://github.com/jegelstaff/formulize/pull/390

PR #390, opened 2023-11-14
Ensuring there are open/close script tags where necessary for non-row subforms

jegelstaff:
A little code tidy up. I think this fixes issue https://github.com/jegelstaff/formulize/issues/389


tikyon (approved):
Fixed variable name. The script is now properly outputted for both non-row types.

PR #391, opened 2023-11-15
Enhancements to the form procedures

jegelstaff:
Adding $newEntry boolean flag to On After Save to indicate first saving event
Adding the ability to return false from On Before Save to abort operation
Making On Delete run after deletion but still have the values of the entry from before deletion available


arvinsingla (commented):
You're missing the change in `writeEntry` function in `data.php` that aborts the save when `false` is returned from `onBeforeSave`

This is what I had in mine.
```
if ($element_values === false) { return null; } 
```

PR #393, opened 2023-11-17
Various style, tidy up, problem fixes, primarily highlighted by VSCode

jegelstaff:


jegelstaff:
@tikyon There is a cleanup detail found here, that might affect the Procedures (advanced calculations). Comment is on the relevant line in entriesdisplay.php

modules/formulize/include/entriesdisplay.php
```diff
@@ -4116,12 +4100,12 @@ function formulize_gatherDataSet($settings, $searches, $sort, $order, $frid, $fi
 		print $GLOBALS['xoopsSecurity']->getTokenHTML();
 	}
 
+	$formulize_cachedDataId = null;
```
jegelstaff, line 4103: @tikyon This variable was not set previously. It is now set, so PHP is happier. But since it had no value, nothing would be sent in $_POST['formulize_cachedata']. That is relevant to advanced_calculations (Procedures) because in the class/advancedCalculation.php file, this value is picked up from POST and used to determine the file name of some cached data. Maybe this should have a value?? Except it hasn't for ages and ages and everything is OK? Since you guys work with the procedures more than anyone, your opinion really matters here.

jegelstaff:
Use ?w=1 on the end of the URL to exclude whitespace changes.

jegelstaff:
Note that there are many changes in class/advancedCalculations.php which should be of no consequence, but some are setting values where previously there would be null or nothing, and maybe it will actually make some things work which were being skipped before? So please check it and try it out to see if there's an effect?

tikyon:
The changes related to advancedCalculation.php looks OK to me. I haven't had a chance to test it on OCANDS yet but it run fine on another project.

On a cleanup note, I notice a lot of "Undefined variable $cellRowAddress" for [entriesdisplay.php L1751](https://github.com/jegelstaff/formulize/blob/ff22d31cef7499ecc96354f50cb02c9003b4eb71/modules/formulize/include/entriesdisplay.php#L1751C155-L1751C170) and [1782](https://github.com/jegelstaff/formulize/blob/ff22d31cef7499ecc96354f50cb02c9003b4eb71/modules/formulize/include/entriesdisplay.php#L1782). This creates several `div#cellcontents__{col}` instead of `div#cellcontents_{row}_{col}`. I think one possible solution is to initialize it before L1736 blocks: `$cellRowAddress = $templateVariables['rowNumber'];`

jegelstaff:
Great catch about the $cellRowAddress. That is in Master now and this has been rebased on that.

tikyon:
When column search is used on list screen, the log shows a lot of "Undefined array key '{handleName}'" for any column not part of the filters on [entriesdisplay.php line 936](https://github.com/jegelstaff/formulize/blob/62a9f1b94b1f33fc56f72261b65a155a8615cff7/modules/formulize/include/entriesdisplay.php#L936). If no column search is used, messages "Undefined variable $searches" for L904,909 and combo of "Undefined variable $searches" & "Trying to access array offset on value of type null" for L936 for each column on the screen.

jegelstaff:
Thanks, $searches is now initialized, and we check isset on the handle before the rest of the code in that block

tikyon:
Thanks the $searches update looks good.

When clicking "Show permissions for these groups" button in form's Permissions tab, the log shows several groups of these two lines: 
> Undefined array key {integer}, modules/formulize/admin/save/form_elements_save.php, 91
foreach() argument must be of type array|object, null given, modules/formulize/admin/save/form_elements_save.php, 91

The number of "Undefined array key" lines seem arbitrary depending on the form. {integer} corresponds to element ID.

tikyon:
When loading the relationship setting page, notice a large number of "Undefined variable $criteria, modules/formulize/class/frameworks.php, 248" in the log.

jegelstaff:
> When loading the relationship setting page, notice a large number of "Undefined variable $criteria, modules/formulize/class/frameworks.php, 248" in the log.

This is fixed in master with 7fe59f5

PR #400, opened 2023-12-04
Database model objects

### LINKED ISSUE #576: New data handling API including query layer, based on data models
Data model PR: https://github.com/jegelstaff/formulize/pull/400

This needs revision to include full query layer, awareness of relationships, presentation of raw value vs readable value. Just in time display of things?

jegelstaff:
This is starting from the database model objects by Daniel Ferguson. Various additional logic based on methods in the data handler could be added, as well as methods in the current form handler and application handler, etc.


jegelstaff:
We need two broad additions to the way this works: 

1. We need a way of flipping 'Formulize Awareness' on and off. This would enable injection of permission filters, scoping, etc, so the API only ever returns what people have access to, like getData does now. This would default to ON. And if you turn it OFF then the API would work like the dataHandler class right now which just works directly with all data.

Note that the dataHandler allows arbitrary scope to be provided in many cases, like groups or user IDs. So perhaps what we need is this on/off flag, and an option to inject certain arbitrary scopes if desired.

2. We need a way of flipping 'Database values' on and off, versus human readable values, which is basically what prepValues does right now.

Or really, perhaps, we need a way of easily giving the user access to both. One query but then when you access the data, we need a way the user can choose which version of the data they're working with.

ie:

```
$entry->element_handle;
$entry->readable('element_handle');
```

jegelstaff:
Also critically, we need an option to retrieve data from the form plus its close relations, ie: one hop away in a relationship, like how getData does now. 

```
$form->plus($relationship)->query()
```

PR #401, opened 2023-12-12
Display improvements and token improvements for inline elements in lists

jegelstaff:


jegelstaff:
This is ready except for testing that the checkbox stuff related to swapping UI text and trans is working as expected (and while we're at it, check on radio buttons too, since they were implicated in the prior code changes)

PR #404, opened 2024-01-08
DOCKER: Improvements for additional functionality

arvinsingla:
* Enables [mod_rewrite](https://httpd.apache.org/docs/current/mod/mod_rewrite.html) in the docker apache instance so we support url rewriting and clean urls.
* Adds libzip and php zip extension to support the formulize synchronization feature
* Sets docker-compose restart policy for services to `no`. This prevents the containers from automatically starting locally when the docker engine started.
* Adding libfreetype and Adding freetype and the freetype php extension
* Adds memory_limit.ini file to increase the base php memory limit to 256M

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -11,9 +11,7 @@ services:
     volumes:
       - .:/var/www/html/
       - ./trust:/var/www/trust
-      - ./docker/php/conf.d/xdebug.ini:/usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
-      - ./docker/php/conf.d/error_reporting.ini:/usr/local/etc/php/conf.d/error_reporting.ini
-      - ./docker/php/conf.d/memory_limit.ini:/usr/local/etc/php/conf.d/memory_limit.ini
+      - ./docker/php/conf.d/formulize.ini:/usr/local/etc/php/conf.d/formulize.ini
     ports:
       - '8080:80'
     links:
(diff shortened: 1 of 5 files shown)


docker-compose.yaml
```diff
@@ -5,21 +5,22 @@ services:
...
-    restart: 'always'
+    restart: 'no'
     depends_on:
       - mariadb
     volumes:
       - .:/var/www/html/
       - ./trust:/var/www/trust
       - ./docker/php/conf.d/xdebug.ini:/usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
       - ./docker/php/conf.d/error_reporting.ini:/usr/local/etc/php/conf.d/error_reporting.ini
+      - ./docker/php/conf.d/memory_limit.ini:/usr/local/etc/php/conf.d/memory_limit.ini
```
jegelstaff, line 16: So my only comment is whether it makes sense to have lots of little .ini files or one big one. Seems to me it depends on the number of customizations we're making? If only a few, why not have just one that is our set of minor changes. If many, then perhaps you want to pick and choose under some circumstances and it's easier to edit which collections are included vs editing the .ini directly? This is really a style question probably? Not sure if there's a significant technical reason to suggest one direction or the other.

jegelstaff:
In general, looks good to me. No qualms about merging this.

docker-compose.yaml
arvinsingla: Good point. There's no technical reason for one over the other so I have no issue with consolidating these into a single ini for our overrides.

PR #411, opened 2024-02-10
Make all code editors use highlighting

jegelstaff:
PHP tags need to be added to the start of all code blocks. And we need to remove the starting tags when we evaluate them.

A function handles removing the tags. A DB patch operation handles adding them. Not having tags is totally OK, but the highlighting won't kick in if you don't put a tag at the top of your code.

The editors that weren't code editors now are code editors thanks to class being added.

forms.php class file has been reorganized a little to be internally consistent in the order of functions.


jegelstaff:
need to make sure derived values have <?php as the value if the ele_value[0] is empty

arvinsingla:
After reviewing this again I'm not convinced it's the approach we should take. Personally I think it's much safer to not require/inject beginning PHP tags on input data and instead inject it at execution time when we know we need it. I get the argument that if a user messes up a code area it's on them, but I think we shouldn't hand them a loaded gun without at least a few safeguards in place.

The primary purpose of this PR is to ensure all code editors use highlighting. This can be accomplished by swapping the `mode` attribute on the CodeMirror instantiation from `application/x-httpd-php` to `text/x-php`.  The use of `x-httpd-php` is good for mixed mode input where you can declare opening/closing PHP tags along with HTML markup which from my understanding is not what we want. Our textareas expect a single PHP code block.

If I'm way off base here please let me know.

modules/formulize/include/functions.php
```diff
@@ -2737,6 +2737,17 @@ function formulize_text_to_hyperlink($text, $textWidth) {
     return str_replace("<a ", "<a target='_blank' ", $text);
 }
 
+/**
+ * Strip <?php out of the beginning of a string if it is present (checks the first five chars)
+ * @param string $string The string you want to check and strip from
+ * @return string The string, potentially modified with <?php gone if it was found as the first five chars
+ */
+function removeOpeningPHPTag($string) {
```
arvinsingla, line 2745: This may be handled elsewhere by the core system but what protection do we have to ensure that no spaces, newline, or invisible characters are placed at the beginning these strings? If any of those characters end up here without a trim in place this function will not strip the PHP tag out.

jegelstaff:
TODO: remove DB patch. Add <?php to necessary places when things are rendering in the admin UI. This will catch all existing cases, and any new elements, new forms, etc, that the user is working with.

modules/formulize/admin/formindex.php
```diff
@@ -609,13 +609,69 @@ function patch40() {
...
+					OR (ele_type IN ('ib', 'areamodif') AND ele_value LIKE '%\$value%')
+					OR (ele_type IN ('text', 'textarea') AND ele_value LIKE '%\$default%') ";
+				if($res = $xoopsDB->query($elementsNeedingOpeningPHPTagsSQL)) {
+					while($record = $xoopsDB->fetchArray($res)) {
+						$eleValueKey = 0;
+						if($record['ele_type'] == 'text') {
+							$eleValueKey = 2;
+						}
+						$newEleValue = unserialize(($record['ele_value']));
+						if(substr($newEleValue[$eleValueKey], 0, 5) != '<?php') {
```
jegelstaff, line 630: Trim the string when comparing

modules/formulize/admin/formindex.php
```diff
@@ -609,13 +609,69 @@ function patch40() {
...
+								print "Notice: could not add opening PHP tag to the code in element ".$record['ele_id']." with the SQL:<br>".str_replace('<', '&lt;',$updateSQL)."<br>".$xoopsDB->error()."<br>This is not a critical error. You can add the tag yourself at the top of the code, if you want the editor to provide highlighting. For more information contact <a href=mailto:info@formulize.org>info@formulize.org</a>.<br>";
+							}
+						}
+					}
+				} else {
+					exit("Error detecting code snippets that need opening PHP tags. SQL dump:<br>".$elementsNeedingOpeningPHPTagsSQL."<br>".$xoopsDB->error()."<br>Please contact <a href=mailto:info@formulize.org>info@formulize.org</a> for assistance.");
+				}
+
+				// Same operation, on the form procedures
+				$formProceduresNeedingOpeningPHPTagsSQL = "SELECT id_form as fid, on_before_save, on_after_save, on_delete, custom_edit_check
```
jegelstaff, line 644: Split into separate operations/functions so it runs once for each event

modules/formulize/admin/formindex.php
```diff
@@ -609,13 +609,69 @@ function patch40() {
...
+					(on_before_save != '' AND on_before_save NOT LIKE '<?php%')
+					OR (on_after_save != '' AND on_after_save NOT LIKE '<?php%')
+					OR (on_delete != '' AND on_delete NOT LIKE '<?php%')
+					OR (custom_edit_check != '' AND custom_edit_check NOT LIKE '<?php%') ";
+				if($res = $xoopsDB->query($formProceduresNeedingOpeningPHPTagsSQL)) {
+					while($record = $xoopsDB->fetchArray($res)) {
+						$events = array('on_before_save', 'on_after_save', 'on_delete', 'custom_edit_check');
+						foreach($events as $i=>$event) {
+							if($record[$event] AND substr($record[$event], 0, 5) != '<?php') {
+								$events[$i] = "$event = ".$xoopsDB->quoteString("<?php\n".$record[$event]);
```
jegelstaff, line 655: Assign trimed version

modules/formulize/admin/formindex.php
```diff
@@ -609,13 +609,69 @@ function patch40() {
...
+					FROM ".$xoopsDB->prefix('formulize_id')." WHERE
+					(on_before_save != '' AND on_before_save NOT LIKE '<?php%')
+					OR (on_after_save != '' AND on_after_save NOT LIKE '<?php%')
+					OR (on_delete != '' AND on_delete NOT LIKE '<?php%')
+					OR (custom_edit_check != '' AND custom_edit_check NOT LIKE '<?php%') ";
+				if($res = $xoopsDB->query($formProceduresNeedingOpeningPHPTagsSQL)) {
+					while($record = $xoopsDB->fetchArray($res)) {
+						$events = array('on_before_save', 'on_after_save', 'on_delete', 'custom_edit_check');
+						foreach($events as $i=>$event) {
+							if($record[$event] AND substr($record[$event], 0, 5) != '<?php') {
```
jegelstaff, line 654: trim the string

modules/formulize/admin/formindex.php
```diff
@@ -609,13 +609,69 @@ function patch40() {
...
+					OR (ele_type IN ('text', 'textarea') AND ele_value LIKE '%\$default%') ";
+				if($res = $xoopsDB->query($elementsNeedingOpeningPHPTagsSQL)) {
+					while($record = $xoopsDB->fetchArray($res)) {
+						$eleValueKey = 0;
+						if($record['ele_type'] == 'text') {
+							$eleValueKey = 2;
+						}
+						$newEleValue = unserialize(($record['ele_value']));
+						if(substr($newEleValue[$eleValueKey], 0, 5) != '<?php') {
+							$newEleValue[$eleValueKey] = "<?php\n".$newEleValue[$eleValueKey];
```
jegelstaff, line 631: trim the assigned text

modules/formulize/admin/formindex.php
jegelstaff: Wouldn't actually simplify things, because the multiple events are stored in the single record, so as written we retrieve the records that need to be updated, and then update the relevant events in the record and write it back to DB. Splitting up would potentially multiply the retrieving and writing by four. Not that we care much in the DB patch operation, but it's needless repetition. Added comments to capture the flow.

PR #413, opened 2024-02-13
Adding conditions for disabling elements

jegelstaff:


jegelstaff:
There's a big more work that could address an edge case where an element has different conditions for display and for disabled. Right now, display conditions would be the only ones respected in that case. Conditional element front end is not designed around differing sets of governing elements for the same underlying element.

jegelstaff:
Todo:
1. refactor saving logic for processing conditions so that everywhere it runs off the parseSubmittedConditions function
2. Reword the intros for display and disabled conditions on the template, so that it references the selected groups, so the user knows they have to pick groups for the settings to take effect

jegelstaff:
Done

PR #414, opened 2024-02-13
Saved views now allow user to choose which aspects are actually being saved

jegelstaff:
TODO: check comments in code, columns and searches need to be made independent. Right now, searches are dependent on being interpreted in parallel with columns.

Also, existing saved views need DB patch to write all features into them so they behave as expected.

Basically, unfinished.


jegelstaff:
Finished now. Needs to be merged. Ready for proper testing by Actua, possibly ESAT tests but not many/any saved view there? Need to check.

jegelstaff:
Also, need to get Dara code into Dara specifically!

PR #415, opened 2024-02-13
Feature/find all entries with all values

jegelstaff:


docs/data-findAllEntriesWithAllValues.md
```diff
@@ -0,0 +1,64 @@
...
+__$fieldsToReturn__ - Optional. the fields to select out of the database. Defaults to _entry_id_. Separate multiple fields with commas, refer to them by their element handles.
+
+## Return Values
+
+Returns __an array of results__, where each item in the array is one record, and each record contains an array where the keys are the field names requested (defaults to _entry_id_) and the values are the values from the database. If _entry_id_ was one of the fields returned, the keys for each record will be the entry ids, otherwise they are sequential from zero.
+
+Returns __false__ if the query fails, or if the query finds no entries that match the values.
```
arvinsingla, line 22: `false` seems overloaded here. The query failing is very different from no entries found that match. Wouldn't returning an empty array be appropriate for no matching values?

Alternatively `false` could serve for no matches, and query failures should result in throwing an exception.

For consistency of the API if other methods use this pattern we can keep it, but maybe some food for thought on how the API could evolve.

arvinsingla (commented):
Overall looks fine, but I have a comment that I'd like to get feedback on before approving.

PR #416, opened 2024-02-18
Updating stylesheet with color variables

jegelstaff:


jegelstaff:
This will make it a lot easier to make simple new themes, just with different colours. One example: https://gvine.ca

themes/Anari/css/style.css
```diff
@@ -10,6 +10,12 @@ body {
   line-height: 1.4;
 }
 
+:root {
```
arvinsingla, line 13: Can we move this to after the meyer reset. Ideally the reset should happen first and then our custom styling afterward.

arvinsingla (approved):
Just one minor comment that I think should be addressed. Otherwise LG2M!

themes/Anari/css/style.css
jegelstaff: Yes!

PR #417, opened 2024-02-18
Adding DESC option for sorting on lists of entries

jegelstaff:


jegelstaff:
The fun part of this, is revising the data storage for the metadata about sorting.

Previously, we were storing a flag of 1 to indicate sorting was happening on a column. And we were assuming ASC, no option for user to choose.

Now, we store ASC or DESC as the flag.

To follow this through from the user perspective, start with the .html template file. Then the .php file in the admin/save folder which reads the submitted values. The submitted values used to just be the index position of the column we want to sort on, ie: 0, 1, 2 for 0th, 1st, 2nd column in the list, etc. Now we pass back 0-ASC or 0-DESC. So we strip the -ASC and -DESC out to check for the index match, and then we store ASC or DESC depending what the submitted value was.

The entriesdisplay.php file reads the stored sorting value out of the DB when the screen loads, and now checks for ASC or DESC and behaves accordingly.

Note that in the .html file, and entriesdisplay.php we are still supporting 1 as the flag for ASC type sorting, because this is the current behaviour so we remain backwards compatible with existing systems, so no DB patching required. But when a save operation happens the value will then be converted to the new format.

PR #421, opened 2024-02-27
A11Y: Changes ported from ESAT

arvinsingla:
Porting various a11y changes from ESAT


modules/formulize/include/functions.php
```diff
@@ -6817,15 +6817,12 @@ class='toggle-button-change-columns'
...
-				$html .= "
-					<div
-						class='elements-checkbox-list'
-						id='id-change-columns-toggle-panel-".$thisFid."'
-						aria-labelledby='id-change-columns-toggle-header-".$thisFid."'
-					>
-						".$boxeshtml."
-					</div>
-				";
+				$html .="<fieldset class='elements-checkbox-fieldset'>";
```
jegelstaff, line 6820: No 'aria-labelledby' in the new version of this, I assume that's on purpose? Just checking.

jegelstaff:
Just one little question, but it looks good to me.

modules/formulize/include/functions.php
arvinsingla: Ahh! Good catch. The aria tag should probably be there.

arvinsingla: Oh wait... I need to double check on this.

arvinsingla: Ahh I see. This was the backport of accessible toggle work I believe.

PR #422, opened 2024-02-27
TASK: No longer delete installation files after install is completed

arvinsingla:
There is no need to delete the installation files once formulize install is completed. This is often laborous for local development where installation is necessary, or for maintaining separate codebases which rebase with master.

This approach will now write an empty `install.lock` file to the file system once installation is completed. All installation files will now check for the presence of a lock file and if one is found will redirect back to the home page. To re-install simply delete the lock file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,4 @@ themes/formulize_mobile_app/.sass-cache/
 formulize-docs/_site
 .DS_Store
 libraries/php-saml/settings.php
-
+install.lock
--- a/install/common.inc.php
+++ b/install/common.inc.php
@@ -43,6 +43,12 @@ icms_Autoloader::setup();
 
 error_reporting( E_ALL );
 
+// Exit early if the system has already been installed
+if (file_exists(ICMS_ROOT_PATH.'/install.lock')) {
+	header('Location: /');
+	exit();
+}
+
 class XoopsInstallWizard {
 
 	var $pages = array();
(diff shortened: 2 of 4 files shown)


jegelstaff (commented):
I think my last two commits solve the issue of when to write the lock file. File is written now after the final DB queries are complete, so site is ready to go. Home link sends user straight to root URL of the site, instead of needing to reload the page_end.php file before redirecting. See what you think.

arvinsingla:
> I think my last two commits solve the issue of when to write the lock file. File is written now after the final DB queries are complete, so site is ready to go. Home link sends user straight to root URL of the site, instead of needing to reload the page_end.php file before redirecting. See what you think.

I had initially tried the same thing but when I did I ended up getting this text on the end page. Even though the file does get successfully created this "failed write file" text appears.

Screenshot 2024-02-28 at 8 58 42 AM

jegelstaff:
Perhaps we should just write the file 'manually'? I've never used that icms thing before. I bet if we do a file_put_contents to create the file it will work. If really just an fopen because we don't have any contents to write.

arvinsingla:
> Perhaps we should just write the file 'manually'? I've never used that icms thing before. I bet if we do a file_put_contents to create the file it will work. Or really just an fopen because we don't have any contents to write.

Ok! I'll give that a try. I was just following suite with how the recurisve delete was doing things, but yeah I'll try a simple approach and see if I get the same results.

PR #423, opened 2024-02-27
Refinements to handling start page settings

jegelstaff:
Anon users now default to Formulize like anyone else. (Due to commit c7ae4bca34af3fd4db1eb31ef2d820afef6021cd)

Problem is that the login block is in the sidebar, which only opens automatically when the user is on the homepage of the site, or on the application.php page in Formulize.

Solution: make the home page detection of whether the user has a start page declared, smarter so it only directs to a Formulize page if the user actually has a start page declared. If they don't stay on home page, and the login block will appear because the sidebar will open.

Typically, anon users won't have a Formulize default page. But now they could thanks to that prior commit, so if they do, great, they get it. If they don't, they stay put and get the login block.

### DIFF
--- a/index.php
+++ b/index.php
@@ -2,7 +2,7 @@
 /**
  * Site index aka home page.
  * redirects to installation, if ImpressCMS is not installed yet
- * 
+ *
  * @copyright	http://www.impresscms.org/ The ImpressCMS Project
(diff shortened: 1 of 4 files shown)


jegelstaff:
TODO: change the Module name in XOOPS from Forms to Formulize!

jegelstaff:
TODO: 
* add return of URL to docblock for code
* Add wrapper methods with clean names instead of passing booleans all around to change method behaviour

PR #424, opened 2024-02-27
FEATURE/One-to-one improvements

jegelstaff:
These are improvements to the conditional element one-to-one handling. These should be merged into the FEATURE/conditional-element-improvements branch, and then that should be merged into the master.


jegelstaff:
Make by mistake really. See FEATURE/conditional-element-improvements

PR #426, opened 2024-02-28
Fix php warning

tikyon:


jegelstaff:
Thanks! I'll check it out.

tikyon:
Some other commonly occurred warnings: 
- Undefined array key "main", modules/formulize/include/extract.php, [Line 736](https://github.com/jegelstaff/formulize/blob/aad3fd5e7f537a9cc679a2f2f4b581882031c6b5/modules/formulize/include/extract.php#L736) 
- Undefined array key "snapshot", modules/formulize/include/extract.php, [Line 1818](https://github.com/jegelstaff/formulize/blob/aad3fd5e7f537a9cc679a2f2f4b581882031c6b5/modules/formulize/include/extract.php#L1818)
- Undefined array key 2, formulize/include/formdisplay.php, [Line 335](https://github.com/jegelstaff/formulize/blob/aad3fd5e7f537a9cc679a2f2f4b581882031c6b5/modules/formulize/include/formdisplay.php#L337)

tikyon:
- Trying to access array offset on value of type bool, modules/formulize/include/extract.php, [Line 942](https://github.com/jegelstaff/formulize/blob/aad3fd5e7f537a9cc679a2f2f4b581882031c6b5/modules/formulize/include/extract.php#L942)
- Undefined variable $entryData, modules/formulize/include/elementdisplay.php, [Line 173-174](https://github.com/jegelstaff/formulize/blob/aad3fd5e7f537a9cc679a2f2f4b581882031c6b5/modules/formulize/include/elementdisplay.php#L173)
- Undefined variable $element_handler, modules/formulize/include/readelements.php, [Line 100](https://github.com/jegelstaff/formulize/blob/master/modules/formulize/include/readelements.php#L100)
- Undefined variable $onetoonekey, modules/formulize/formulize_xhr_responder.php, Lines 202, 210, 236 (grouped)

jegelstaff:
Terrific, love to clean this stuff up. We have altered some of the assignments, so instead of ?? they use the more common ? ternary operator. This is because ?? only does its thing when the value is null, and the ternary will catch anything that evaluates to false, just to be more broad and catch any possible invalid values.

Also, reverted the change in elementdisplay.php since upcoming code in the master branch will conflict with that. The entire operation is now cleaner and more sensible, or will be shortly.

tikyon:
Just a note that [1b43c99](https://github.com/jegelstaff/formulize/pull/426/commits/1b43c99ffc84cf8b8568c892495a46e25458cd8c) brings back the warning "Undefined global variable $formulize_screenCurrentlyRendering"

jegelstaff:
Needs to be an (isset and true) comparison to avoid the undefined variable issue. Easy to resolve. The others should possibly, probably do the same.

tikyon:
The same assignment `$previouslyRenderingScreen = $GLOBALS['formulize_screenCurrentlyRendering']` is also in 

- modules/formulize/class/calendarScreen.php Line 169
- modules/formulize/class/multiPageScreen.php Line 240

jegelstaff:
Finally fixed all warnings in https://github.com/jegelstaff/formulize/commit/baef59f6ac2a0f92b943b1ff89dfea3d40588f23

PR #429, opened 2024-03-07
Feature/smarter conditional element catalogue

jegelstaff:
Also, moves the conditional logic js to its own file


jegelstaff:
Todo:
* Test on Massey

jegelstaff:
This has interaction effects with the new disabled conditions, which should be finished first, and then this updated to match.

PR #433, opened 2024-03-20
Empty conditional HTML Capture dom element after capturing HTML. Duh!

jegelstaff:


modules/formulize/include/formdisplay.php
```diff
@@ -4318,7 +4318,9 @@ function callCheckCondition(name) {
 function captureDataAsInDOM(data) {
 	jQuery('#conditionalHTMLCapture').empty();
 	jQuery('#conditionalHTMLCapture').append(data);
-	return window.document.getElementById('conditionalHTMLCapture').innerHTML.trim();
+	var capturedHTML = window.document.getElementById('conditionalHTMLCapture').innerHTML.trim();
```
arvinsingla, line 4321: Moving forward I would recommend shying away from the use of `var` in javascript. Whenever possible use `const`, followed by `let` if mutation is required. The primary purpose is to prevent JS variable scoping issues as a best practice. All browsers have supported const/let for a long time now

modules/formulize/include/formdisplay.php
```diff
@@ -4318,7 +4318,9 @@ function callCheckCondition(name) {
 function captureDataAsInDOM(data) {
```
arvinsingla, line 4318: I'm confused by the actual purpose of this function now. If you're ultimately emptying it then what's the point?

arvinsingla (changes requested):
I'm not sure I understand the purpose of the `captureDataAsInDOM` function given the updates.

modules/formulize/include/formdisplay.php
jegelstaff: The purpose is to ensure that when we are comparing a set of markup we received from the server against the current state of an element in the form, the markup will be formatted exactly the same.

All the markup in the form for the conditional elements is captured when the form is loaded. It's captured by pulling it out of the DOM with innerHTML. There is a global array into which it is put. 

And when conditions fire, we compare what we get back from the server against that array, to see if we actually need to animate anything. Because there might have been a change to the governing elements and that fired a condition, but it did not actually result in a change to the conditional element. So getting a result back from the server isn't enough to tell us whether we should animate anything in the UI. We have to check for changes.

The problem is that if we compare the string we get back from the server to the markup that we pulled out of the DOM when we catalogued everything, they can be structurally the same but still differ in terms of quoting or other strange things that JavaScript and/or the browser does to the markup when it goes in and out of the DOM.

So if you want to take a piece of raw markup from the server and do a string comparison against the markup as it exists in the DOM currently, the only way to do that reliably is to shove the markup from the server into the DOM and pull it back out again. Because then it has gone through exactly the process that the markup for the current state of the element went through. It's the only way of ensuring a proper apples to apples comparison.

Strange but true.

arvinsingla: Thanks for the explanation. I would almost certainly classify that as a hack which means there should probably be some inline documentation associated with it.

Maybe just add some documentation and we can consider this 👍🏾

arvinsingla (approved):

PR #435, opened 2024-03-28
Keeping notification queue history for a week

jegelstaff:
Instead of deleting the notification queue when it is all processed, move the file into a queue_history folder. Keep up to one week of files in the folder.


jegelstaff:
This has not been formally tested, though the change seems obvious and clean enough. Validating by deploying to dev or staging on esat, and checking if it moves a file to the right folder, would be a wise step.

jegelstaff:
Thanks @arvinsingla for the discussion and suggestion!

PR #438, opened 2024-03-31
Various Grid element improvements

jegelstaff:


jegelstaff:
Grids are an ancient solution to the problem of how to show elements side by side. They basically create a xoopsFormLabel where the contents side is an HTML table that contains the elements. So... first thing, table should be trashed in favour of divs, etc.

But setting that aside, this improves the layout, because the row captions and column captions are only shown if there are any. You might specify empty text for the captions and that should be OK and make the layout look clean.

Also, super important, grids are conditional now. You can show/not show like any other element. Needs big refactoring!

Also, grids will show the required symbol if any element in the grid is required.

Grids determine their elements by starting with the specified element chosen in the element options, and selecting the next X elements in order to fill up all the spaces in the grid (rows x columns).

There is maybe a future where grids are a key part of an advanced form layout system, where you can use the GUI to drag and drop elements, move them into grids, etc. This functionality is super handy, usually just for putting two or three elements in a row (day, month, year, or city province, etc). And it's way too hard to access right now. A refactoring of the code, so there's a grid element class, that had stripped down options... that is probably the way to go... but in a GUI form editor, you don't want users to have to specify it as an element. The grid is just a quirk for controlling element layout. But then it also needs to be an element in the sense that it has an order and position, and can be chosen to include in a screen on a page, etc. So a lot of subtlety here, but in general would be nice if the way you access this functionality were simplified. Going to that level is beyond the scope of this PR, which is touching up quality of life things, and the conditional integration is huge. We get that it and then Formulize 9 can have big element management improvements?

jegelstaff:
Todo: need to ensure elements that are part of a conditionally hidden grid are not displayed on their own. They need to be in a form so they appear on mobile because grids don't show on mobile. Or something like that. This needs testing.

jegelstaff:
image

jegelstaff:
Last thing: validation JS for required elements inside grid needs to be wrapped in conditional safeguards, if the element has conditions on it. Making a required grid element disappear means that the validation JS fails because it can't find its target any longer.

jegelstaff:
Also, can prevEntry be put into loadValue? Refactor out of where it's generated now?

jegelstaff:
There is one edge case that we don't handle cleanly... conditionally disabled elements that are also required. They work inside grids (maybe?). Definitely don't work outside grids, if they start out disabled, and become enabled. The required validation js does not kick in under those circumstances.

This is captured in https://github.com/jegelstaff/formulize/issues/458

PR #439, opened 2024-03-31
Show thumbnails instead of text if uploaded file is an image

jegelstaff:


jegelstaff:
Kinda bold to just show the image, rather than making a thumbnail version. What if the image is super large? But is that really more on the user? Possibly an issue worth fixing when it's an issue. Because the functionality improvement of this approach is huge and is worth getting into place.

Also not sure if setting 200px max width is the best choice to control the size (if we're not making thumbnails). Does the trick. Again, may be a case of further optimizations later when it's warranted?

jegelstaff:
See https://github.com/jegelstaff/formulize/issues/445 and https://github.com/jegelstaff/formulize/issues/446 for enhancement options.

PR #440, opened 2024-03-31
Make multipage form tabs fixed position

jegelstaff:


jegelstaff:
Also fixes git snafu that got two versions of colour variables into the CSS!

jegelstaff:
On mobile, if the tabs wrap, the top of the form should be lower... but how to figure out where the bottom of the fixed position div is?? If the spacer div with the nbsp had a line break then it would space out to match the wrapping of the tabs, but again, how do we know when this is necessary??

jegelstaff:
JS in place. Ready for review. Question is, should the JS look like it does now? Or should it look like it did here: https://github.com/jegelstaff/formulize/pull/440/commits/a9f185085cf878ba74d5a4868a661fe074f2100c

jegelstaff:
Let's investigate 'sticky' !!

PR #442, opened 2024-03-31
Range slider improvements

jegelstaff:
Bigger text, easier to read. Matching width to autocomplete elements when full width, etc.

Making sliders trigger conditions properly.


jegelstaff:
Todo: actually delete the newslider files in the patch process, because having them around will allow people to make newslider elements!

jegelstaff:
Todo: when going to a subform entry newly created with the Add button in the subform interface, sliders don't initialize properly.

PR #444, opened 2024-04-01
Further module name updates

jegelstaff:
Update in DB, so user doesn't have to type a new name.


jegelstaff:
We also delete the cached menu on the admin side since it will have set things up based on the old name.

PR #451, opened 2024-04-05
Use method in custom element class, if one exists

jegelstaff:


jegelstaff:
This is important because this function is used in the saving of data, to determine what the default values should be for any entries that are 'new.' So if an element isn't represented here, then no defaults are possible in that sense (default values would still show up when the element appears on screen as 'new' but if the element isn't on screen when the entry is 'new' then no default value is saved).

With the connection to the element type classes, this way the element class files can do their own thing regarding default values and everything will "just work."

PR #452, opened 2024-04-05
Use proper function to write new subform entries, causes defaults to …

jegelstaff:
…behave more as expected


jegelstaff:
New subform entries were one of the last places where the old writeElementValue function was being used. This now switches the writing process to the correct part of the current API, and it simplifies things since the API allows writing an entire record at once, instead of writing a single value like the ancient function this replaces.

Note the unusual repeated application of defaults after the entry is written. This is a cheat to allow date elements that reference a start date in another element in the entry, to pick up that value after saving, and then update/rewrite the defaults again.

PR #453, opened 2024-04-05
Apply quotes only to non numerics when processing display and disable…

jegelstaff:
…d conditions

### DIFF
--- a/modules/formulize/include/elementdisplay.php
+++ b/modules/formulize/include/elementdisplay.php
@@ -479,25 +479,30 @@ function buildEvaluationCondition($match,$indexes,$filterElements,$filterOps,$fi
 			$compValue = implode(",",$compValue);
 		} else {
 			$compValue = addslashes($compValue);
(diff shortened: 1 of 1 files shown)


jegelstaff:
This PHP level parsing of conditions is meant to work nicely with data retrieved from the DB, ie: get an entry in the normal way, and then we use PHP to see if it meets the conditions.

This predates the ability to generate clean SQL from the conditions, using the complex functions that handle that operation. We could refactor this to use the other approach, which might be more powerful, cover more bases. But this has been working and that would be a very big change, testing existing conditions to ensure they work the other way would be paramount. Not worth changing yet.

Also, the SQL way generates a complex query that might take a little while to play out, whereas getting a single entry by ID is super quick. So this may have performance benefits, though I expect they are minor.

One benefit of doing something based on SQL, is that those functions can handle a query that involves aliased additional forms in a relationship, so filters could be based on more than just the questions in this one form. That would be pretty wild and complex, but theoretically possible.

PR #462, opened 2024-04-10
Moving conditionalHTML markup to render one time when conditional.js is injected

arvinsingla:
This colocates the rendering of the conditionalHTMLCapture div alongside the inclusion of conditional.js as they are implicitely linked.


jegelstaff:
Looks good to me. This is basically exactly what I did in production... Copy paste that line to right after the JS. :)

PR #463, opened 2024-04-12
If a caption has { } references, make the element conditional

jegelstaff:
Also cleans up variable name throughout.


jegelstaff:
This way if you have a dynamic caption, you don't also have to mark the element as conditional dependent on the other element. It just gets conditional behaviour automatically.

arvinsingla (approved):
Reviewed this with Julian directly. LG2M 👍🏾

PR #464, opened 2024-04-12
If admin key matches a system element, only skip saving if we're saving an element!

jegelstaff:
Anon passcodes are system elements. Part of a form but not editable by users. We have a block on saving elements if they are system elements. It was mistakenly operating at too high a level and would bail on saving if the item we were saving happened to have an id that matched the id of a system element (ie: element 22 is a system element, we're saving form 22, so nothing in form 22 saves!)


arvinsingla (approved):
Discussed with Julian directly. LG2M 👍🏾

PR #465, opened 2024-04-12
Ignore values that haven't changed or won't change when synching

jegelstaff:
"" and 0 amount to the same thing, if the field is a kind of integer field. 

And if the value to be imported, is the same as the value in the DB after being corrected for the groups-in-common, then we don't need to actually import it.

And tidy up strings in the JS alerts, so they don't break so often.


jegelstaff:
@tikyon If you are able, could you try out this branch with a real synchronization? I believe the changes are good and correct, they have been used in another live system. But I would appreciate verification from another source.

tikyon:
I tested and didn't run into any issues.

jegelstaff:
> I tested and didn't run into any issues.

Thank you!!

PR #468, opened 2024-04-19
Fix/lower case table names in docker

jegelstaff:


jegelstaff:
So, for the record, there is a serious issue when taking a phpMyAdmin dump from Linux and trying to replicate the DB locally in Docker on Windows, given the Formulize setup. Details are here: https://mariadb.com/kb/en/mariadb-containers-should-mariadb-docker-containers-really-be-using-macos-s/

This PR provides a solution, by adding an optional .cnf file for MariaDB to the repository. The developer has to manually enact this file in their system if they want to use it. Doing so should not affect the repository, because the .cnf file they create/use will be ignored by Git, so our clean default distribution that does not use this, will be unaffected, but Windows based developers who need this fix can enact it and life will go on.

PR #472, opened 2024-04-20
Header row changed to th, updates needed to keep sticky headers working in LOE

jegelstaff:


jegelstaff:
This was a side effect of accessiblility fixes I think

PR #474, opened 2024-04-21
Making direct limit restriction on linked selectboxes aware of condit…

jegelstaff:
…ions


jegelstaff:
Imagine a list of authors, linked from a master list of people. Imagine a primary author element also linked to the master list. In the entry for a single publication, the primary author list should not be all available people, it should only be the list of authors of the publication (pick one as primary). This is accomplished with a direct limit on the entries included (limit them based on the values in element X).

Well if element X is the governing element for a display condition on the other element that has the direct limit, this code lets the connection behave as expected. Changing the author list, changes the available primary authors in real time.

The asynch data from the condition needs to be respected when generating the options for the element with the direct limit, since the new selection is not yet in the DB.

PR #483, opened 2024-04-24
Get grid elements based on grid count alone

jegelstaff:
This is kind of important now. In the beginning, no conditional elements, and grids were designed to always be full, and if an element wasn't displayed to a user, we kept going till we found an element in the form that the user could see, and the grids were always full. This is kind of dumb when there are conditional elements and other intricate ways of organizing a form now. There are certain elements meant for the grid, and other elements are absolutely not meant for the grid!

So grids always get their elements based on count and only display those and that's it. This also created problems when dealing with the assignment of grid conditions to the constituent elements, and so doing things this way keeps stuff nicely inside the grid only.


jegelstaff:
This seems to have got too many commits, remaking...

PR #492, opened 2024-04-25
Introduce new e2e testing scaffolding

arvinsingla:
This PR ads the foundation of using Playwright as the new e2e test runner.

The PR ads the following:
* `tests` folder which will house all tests moving forward (e.g. unit and e2e)
* `e2e` folder inside `tests` which adds the package.json, playwright config, and test files
* github action to execute the e2e tests. Currently only set to run when files are merged into master

**Q&A**

**Why not use the existing CI folder for tests?**
CI is a bit of a misnomer. CI specific configuration files are now stored in their namespaced directories (e.g. `.github`). So `/tests` is where we would expect the tests themselves to live. Since these tests can be run in many environments (e.g. CI, Locally, etc) it's confusing if they stay in the CI folder

**Why keep the CI folder?**
Once all the existing tests have been ported over to playwright we can remove the CI folder. Until then they are still available and can be used.

**Why are the saucelabs steps commented out in the github actions file?**
I am still evaluating the usefullness of using SauceLabs vs. having playwright generating testing output locally. Currently playwright is generating the testing output and writing it as assets to the action.


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,57 @@
+name: e2e test suite
+
+on: [push]
```
arvinsingla, line 3: This will be changed to pushes to `master` prior to being merged in. It's currently set to on push to demonstrate it's functionality.

jegelstaff: So... continuous integration... but do we really need all this operating after every commit? Or only after.... ?? It's tidier to say every commit. But if you're fixing typos... don't need the tests to run. Maybe we should have a reverse flag, [NO TEST] or something like that, similar to the [DEPLOY TO DEV] in the specific site repos? ie: a way to skip the tests when it's a waste of time and resources.

jegelstaff:
In general, looks great. Just see the one comment about whether we can have a way to turn this off if we really don't need it on a commit/merge

.github/workflows/e2e-test.yml
arvinsingla: Yeah, I could definitely add a skip flag to prevent the tests from being run on a given commit. This will also evolve over time so we can see how it plays out. The ideal case in the future is to have the tests run at the PR stage and prevent merging if tests fail, but that's a future state.

arvinsingla:
Bases on updates this is now RTBC

PR #493, opened 2024-04-27
Improvements to handling anon scenarios, sending them to login

jegelstaff:


jegelstaff:
Related, think about the login block being on every page? But for now keeping it on top page and booting people to the homepage seems reasonable. Login block has to be accessed by opening menu, except on front page where it opens automatically. So booting to front is comprehensive.

PR #503, opened 2024-05-07
Always load some session if its there, never nullify an existing session!

jegelstaff:
In strange cases (multiple requests to server at once from same user, because of XHR or img tags, etc), we might loop through the entire set of ticks when reading the session, without slipping through to the actual loading of the session. In that case, return empty string. Ack! No more session!

Now when the loop is done, we go with whatever session we have found, no matter what. Session is only nullified if there's a change in IP address, which is a totally valid security measure against session stealing.


jegelstaff:
Also, instead of waiting a second between checks, we wait 1/3rd of a second, which may reduce the waiting time, without massively increasing the burden on the DB to be checking the session table.

PR #508, opened 2024-06-10
Feature/logfiles

jegelstaff:
Log files are written one per day. New lines appended to the end with new logged events. Format is json.

Single function call to write to the log. Key-Value pairs are passed in. Additional core metadata included with each line.

Log location and storage duration are controllable through Formulize preferences.


jegelstaff:
Should file extension be .json ?

Separator of lines is the new line character. Full JSON wouldn't have that?

modules/formulize/include/functions.php
```diff
@@ -8051,3 +8073,50 @@ function generateSelfReferenceExclusionSQL($entry_id, $fid, $sourceFid, $ele_val
...
+ *
+ * @param array $data Key-Value pairs that should be written to the log entry
+ * @return int|boolean Returns the number of bytes written to the file, or false on failure
+ */
+function writeToFormulizeLog($data) {
+
+	// initialize the configuration settings
+	static $formulizeConfig = false;
+	static $formulizeLogFileLocation = XOOPS_ROOT_PATH.'/logs';
+	static $formulizeLogFileStorageDuration = 168;
```
arvinsingla, line 8088: To avoid ambiguity I might suffix with the unit of measurement. 
```suggestion
	static $formulizeLogFileStorageDurationHours = 168;
```

modules/formulize/include/functions.php
```diff
@@ -8051,3 +8073,50 @@ function generateSelfReferenceExclusionSQL($entry_id, $fid, $sourceFid, $ele_val
...
+/**
+ * Record information in a log file. One log file per day. Log file location is a Formulize preference. Log file storage duration is a Formulize preference.
+ *
+ * @param array $data Key-Value pairs that should be written to the log entry
+ * @return int|boolean Returns the number of bytes written to the file, or false on failure
+ */
+function writeToFormulizeLog($data) {
+
+	// initialize the configuration settings
+	static $formulizeConfig = false;
```
arvinsingla, line 8086: Just so I understand correctly. The use of `static` in this context is to ensure that when the function exits the value is retained so any additional call to the function in the same request will use those existing values?

modules/formulize/include/functions.php
```diff
@@ -8051,3 +8073,50 @@ function generateSelfReferenceExclusionSQL($entry_id, $fid, $sourceFid, $ele_val
...
+	if(!$logFilesCleanedUp) {
+		formulize_scandirAndClean($formulizeLogFileLocation, 'formulize_log_', $formulizeLogFileStorageDuration * 60 * 60);
+		$logFilesCleanedUp = true;
+	}
+
+	// prepend certain metadata to the data we're logging
+	// UNIQUE ID is only available if the server has mod_unique_id turned on
+	$data = array(
+		'microtime' => microtime(true),
+		'request_id' => (isset($_SERVER['UNIQUE_ID']) ? $_SERVER['UNIQUE_ID'] : 'mod_unique_id is not enabled'),
```
arvinsingla, line 8110: The fallback case here won't fly with the log processing. I think we should fallback to the microtime in this case.

modules/formulize/language/french/modinfo.php
```diff
@@ -66,3 +66,8 @@
 
 define("_MI_formulize_USEOLDCUSTOMBUTTONEFFECTWRITING", "Use the old method of writing effects for custom buttons");
 define("_MI_formulize_USEOLDCUSTOMBUTTONEFFECTWRITINGDESC", "This should always be \"No\" unless this is an older installation that already has custom buttons that are dependent on the old method, which was based on the declaring human readable values, instead of the database values for elements.");
+
+define("_MI_formulize_FORMULIZELOGFILELOCATION", "Location to store Formulize log files");
+define("_MI_formulize_FORMULIZELOGFILELOCATIONDESC", "Formulize generates log files that contain the history of user actions, such as logging in and saving data. You can specify the full path to the folder where the log files are stored. You can turn off logging by specifying an empty or non-existent path.");
```
arvinsingla, line 71: Feels a little clunky that there isn't an explicit on/off for the logs. Especially since you're putting a default value which implies that logging is "enabled" but only so long as the directory exists.

arvinsingla:
> Should file extension be .json ?
> 
> Separator of lines is the new line character. Full JSON wouldn't have that?

The log file itself is not valid JSON so it should not be a `.json` extension. It's a log file where each log line is a JSON string.

jegelstaff:
We need to standardize the fields in the log lines, so that all log lines are equivalent, so that processing and ingestion is standardized. Some lines will have empty values when the field is not applicable to that operation that is being logged.

modules/formulize/include/functions.php
```diff
@@ -8051,3 +8073,50 @@ function generateSelfReferenceExclusionSQL($entry_id, $fid, $sourceFid, $ele_val
...
+		'session_id' => session_id()
+		) + $data;
+
+	// write the new log entry (to a new file if necessary, active file has generic name, archived files are named with the current date based on server timezone)
+	// write operation is self contained in file_put_contents and closes the file, so it's available for other concurrent requests to write to after
+	$logFileName = 'formulize_log_'.date('Y-m-d').'.log';
+	$activeLogFileName = 'formulize_log_today.log';
```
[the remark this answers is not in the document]
jegelstaff: Change to formulize_log_active

modules/formulize/include/functions.php
arvinsingla: Upon further reflection we could fallback to https://www.php.net/manual/en/function.uniqid.php

jegelstaff: We will generate a PHP unique id as a static variable and reuse that, if there is no value for $_SERVER['unique_id']

modules/formulize/language/french/modinfo.php
jegelstaff: Add on/off preference

jegelstaff:
@arvinsingla Changes done. This is ready for trying out now.

PR #514, opened 2024-06-26
Bug fix for resynch of changed checkbox options

jegelstaff:


jegelstaff:
Checkbox class needs to handle the resynch because it handles the processing of the ele_value values into key 2 now!

Also, some sloppy typos referring to element type 'check' instead of 'checkbox' and the separator string was wrong for checkboxes and selects!!

PR #538, opened 2024-09-07
Entry identifiers with clean URLs

jegelstaff:
This adds the option to select a field in a form, that is used to generate an identifier for the entry, when clean URLs are being used. Must be set on both form and list / template and list screens in order to work as expected.


jegelstaff:
A work in progress. Still needs things worked out in formdisplay.php. Pay attention to comment there.

jegelstaff:
Initially tested out on Massey.

jegelstaff:
Work for regular forms now. And start page redirects. But you have to use the specific screen as the start page. Does not work for templates that are including forms, and other funky template stuff. But could maybe in the future?

PR #541, opened 2024-09-19
Enable PHP syntax checking

jegelstaff:
This operation can be enabled/disabled through a preference. It takes time to run, and runs with every save operation. This may be inconvenient for some developers/webmasters.

This operation will only take effect if the server has shell_exec enabled for PHP.

This code removes the xDebug reference in the formulize.ini that is used in Docker. This appears to be unnecessary because the dockerfile that controls our PHP instance already includes/enables xDebug? With the declaration in formulize.ini in place, there is an error on PHP start up, and that shows up as an error in every code block. With the declaration removed, this extra start up error is not present, and xDebug functionality in Docker/VSCode still works like normal.

Textarea boxes that have the class canValidate, are catalogued by the JS code, and then sent to the server which does the shell_exec PHP syntax check. Nothing returned (empty string) means there's no problems. If problems are found, then they are sent to the user in a JS alert and the saving operation does not proceed.

The UI fades out as if saving, while the check is taking place. The "You have unsaved changes" warning does not go away if the syntax check fails. Instead the screen just returns to full brightness and the user can continue from there.

### DIFF
--- a/docker/php/conf.d/formulize.ini
+++ b/docker/php/conf.d/formulize.ini
@@ -4,9 +4,6 @@ error_reporting=E_ALL
 ; Memory Limit
 memory_limit = 256M
 
-; XDebug
-zend_extension=xdebug
-
 [xdebug]
 xdebug.mode=develop,debug
 xdebug.discover_client_host=true
--- a/modules/formulize/libraries/formulize-admin.js
+++ b/modules/formulize/libraries/formulize-admin.js
@@ -18,25 +18,25 @@ jQuery(document).ready(function() {
 
 });
 
-async function fz_check_php_code(custom_code, block_name) {
+function fz_check_php_code(custom_code) {
 	let validateCode = new Promise(function(resolve, reject) {
     jQuery.ajax({
         type: "POST",
         url: window.icms_url+"/modules/formulize/formulize_xhr_responder.php?uid="+window.icms_userid+"&op=validate_php_code",
         data: {the_code: custom_code},
         success: function(result) {
-            if (result.length > 0) {
-              resolve("The "+block_name+" has an error:\n\n"+result+".");
-            } else {
-							resolve('');
-						}
+							resolve({
+								valid: result.length > 0 ? false : true,
+								result: result
+							})
         	},
 				error: function() {
-					reject('The validation request to the server failed.');
+					reject({
+						valid: false,
+						result: 'The validation request to the server failed.'
+					})
 				}
     });
 	});
-	return validateCode.then(result => {
-		return result;
-	});
+	return validateCode;
 }
--- a/modules/formulize/libraries/formulize-admin.js
+++ b/modules/formulize/libraries/formulize-admin.js
@@ -15,19 +15,28 @@ jQuery(document).ready(function() {
         }
         hidden_section.toggle();
     });
-    
+
 });
 
-function fz_check_php_code(custom_code, block_name, icms_url, icms_userid) {
+async function fz_check_php_code(custom_code, block_name) {
+	let validateCode = new Promise(function(resolve, reject) {
     jQuery.ajax({
         type: "POST",
-        url: icms_url+"/modules/formulize/formulize_xhr_responder.php?uid="+icms_userid+"&op=validate_php_code",
+        url: window.icms_url+"/modules/formulize/formulize_xhr_responder.php?uid="+window.icms_userid+"&op=validate_php_code",
         data: {the_code: custom_code},
         success: function(result) {
             if (result.length > 0) {
-                alert("The "+block_name+" has an error:\n\n"+result+".");
-            }
-        },
-        async: false
+              resolve("The "+block_name+" has an error:\n\n"+result+".");
+            } else {
+							resolve('');
+						}
+        	},
+				error: function() {
+					reject('The validation request to the server failed.');
+				}
     });
+	});
+	return validateCode.then(result => {
+		return result;
+	});
 }
--- a/modules/formulize/libraries/formulize-admin.js
+++ b/modules/formulize/libraries/formulize-admin.js
@@ -27,7 +27,7 @@ function fz_check_php_code(custom_code) {
         success: function(result) {
 							resolve({
 								valid: result.length > 0 ? false : true,
-								result: result
+								result
 							})
         	},
 				error: function() {
--- a/modules/formulize/templates/admin/ui.html
+++ b/modules/formulize/templates/admin/ui.html
@@ -181,15 +181,19 @@ $(document).ready(function() {
 	async function validateCode() {
 		let codeBlocks = [];
 		let error = '';
+		let errorMsg = '';
 		$(".canValidate").each(function() {
 			codeBlocks.push({ code: $(this).val(), title: $(this).prevAll('legend').first().text() });
 		});
 		for (let i = 0; i < codeBlocks.length; i++) {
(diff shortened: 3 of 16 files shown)


modules/formulize/templates/admin/ui.html
```diff
@@ -1,309 +1,335 @@
...
+	});
+
+	async function validateCode() {
+		let codeBlocks = [];
+		let error = '';
+		$(".canValidate").each(function() {
+			codeBlocks.push({ code: $(this).val(), title: $(this).prevAll('legend').first().text() });
+		});
+		for (let i = 0; i < codeBlocks.length; i++) {
+			error = await fz_check_php_code(codeBlocks[i].code, codeBlocks[i].title);
```
jegelstaff, line 188: Note that you cannot reference async functions with await, inside a closure. Otherwise we would do this inside the .each operation above. But that's too wild and crazy, even for JS. So we have to catalogue the blocks, and then call the actual async function that sends off the ajax requests, inside a very ho-hum for loop.

docker/php/conf.d/formulize.ini
```diff
@@ -4,9 +4,6 @@ error_reporting=E_ALL
 ; Memory Limit
 memory_limit = 256M
 
-; XDebug
```
jegelstaff, line 7: As mentioned in the lead comment, this code removes the xDebug reference in the formulize.ini that is used in Docker. This appears to be unnecessary because the dockerfile that controls our PHP instance already includes/enables xDebug? With the declaration in formulize.ini in place, there is an error on PHP start up, and that shows up as an error in every code block. With the declaration removed, this extra start up error is not present, and xDebug functionality in Docker/VSCode still works like normal.

jegelstaff: This needs verification on another system besides my own!

arvinsingla: I've confirmed that when the container is built we do the following `RUN docker-php-ext-enable mysqli pdo pdo_mysql gd xdebug zip` which is functionally equivalent to adding `zend_extension=xdebug` in the .ini file. Additionally I have built a version of this image and used it with success in running xdebug. This should be safe to remove.

modules/formulize/libraries/formulize-admin.js
```diff
@@ -15,19 +15,28 @@ jQuery(document).ready(function() {
         }
         hidden_section.toggle();
     });
-    
+
 });
 
-function fz_check_php_code(custom_code, block_name, icms_url, icms_userid) {
+async function fz_check_php_code(custom_code, block_name) {
```
arvinsingla, line 21: Since this function itself does not make use of the `await` keyword there's no need to call it an async function. 
```suggestion
function fz_check_php_code(custom_code, block_name) {
```

modules/formulize/libraries/formulize-admin.js
```diff
@@ -15,19 +15,28 @@ jQuery(document).ready(function() {
...
+						}
+        	},
+				error: function() {
+					reject('The validation request to the server failed.');
+				}
     });
+	});
+	return validateCode.then(result => {
+		return result;
+	});
```
arvinsingla, line 41: This should not be necessary. What you are sending in the resolve method will be the value returned when the promise is resolved.

modules/formulize/libraries/formulize-admin.js
```diff
@@ -15,19 +15,28 @@ jQuery(document).ready(function() {
...
         success: function(result) {
             if (result.length > 0) {
-                alert("The "+block_name+" has an error:\n\n"+result+".");
-            }
-        },
-        async: false
+              resolve("The "+block_name+" has an error:\n\n"+result+".");
+            } else {
+							resolve('');
+						}
```
arvinsingla, line 32: The purpose of this function is to perform a validation on the given inputs and return the results. IMHO this method should not perform UX tasks like calling alert or responding with strings that are destined for end user consumption. That should be the responsibility of the calling method.

Additionally I believe the response data from this method should be an object and not a string. 

```
resolve({
  valid: result.length > 0 : false : true
  result
})
```

arvinsingla (changes requested):
Good first pass, I have some comments I think we should address.

modules/formulize/libraries/formulize-admin.js
jegelstaff: Confirmed. Thanks!

modules/formulize/libraries/formulize-admin.js
jegelstaff: Confirmed. Thanks!

modules/formulize/libraries/formulize-admin.js
jegelstaff: This is addressed in the latest commit. Thanks!

jegelstaff: @arvinsingla Take a peek pls

modules/formulize/libraries/formulize-admin.js
```diff
@@ -15,19 +15,28 @@ jQuery(document).ready(function() {
...
         data: {the_code: custom_code},
         success: function(result) {
-            if (result.length > 0) {
-                alert("The "+block_name+" has an error:\n\n"+result+".");
-            }
-        },
-        async: false
+							resolve({
+								valid: result.length > 0 ? false : true,
+								result: result
```
arvinsingla, line 30: Suggestion: `result: result` is redundant in JS. If the key and the variable are the same you can just use the following. 

```suggestion
								result
```

modules/formulize/templates/admin/ui.html
```diff
@@ -1,309 +1,339 @@
...
+	async function validateCode() {
+		let codeBlocks = [];
+		let error = '';
+		let errorMsg = '';
+		$(".canValidate").each(function() {
+			codeBlocks.push({ code: $(this).val(), title: $(this).prevAll('legend').first().text() });
+		});
+		for (let i = 0; i < codeBlocks.length; i++) {
+			error = await fz_check_php_code(codeBlocks[i].code);
+			if(error.valid == false) {
```
arvinsingla, line 190: Suggestion: Usually best practice to perform a strict equality check.

```suggestion
			if(error.valid === false) {
```

modules/formulize/templates/admin/ui.html
```diff
@@ -1,309 +1,339 @@
...
+				errorMsg = "The "+codeBlocks[i].title+" has an error:\n\n"+error.result+".";
```
arvinsingla, line 191: Suggestion: Template strings remove the need for ugly string concatenation. 

```suggestion
				errorMsg = `The ${codeBlocks[i].title} has an error:\n\n${error.result}`;
```

arvinsingla (approved):
👍🏾 I have not functionally tested this feature so if you have and it looks good.  I have a few minor suggestions but none are blocking approval.

jegelstaff:
Suggestions look good. All committed.

Further to do: only check if there is code in the block to check. Otherwise don't send time consuming ajax.post request.

jegelstaff:
In general, the caching feature works because each page of the admin UI is loaded into the browser on its own. We do not have a persistent JS based admin UI where everything is dynamic and it's one JS session/scope happening start to finish. Every page the user visits is a new page, new http request. So the JS functions only on that page. Therefore the checkedBlocks array is the set of checked code blocks *for that page only*. Keeps the scale down of what we're storing in memory because it is flushed every page load, and it means we don't need something more elaborate for tracking such changes.

PR #544, opened 2024-09-24
Primary Relationship

jegelstaff:
See related issue(s)


jegelstaff:
This needs to handle empty/new systems, where there are no forms, no connections. We should setup the Primary Relationship, with no links, as part of the install.

jegelstaff:
> This needs to handle empty/new systems, where there are no forms, no connections. We should setup the Primary Relationship, with no links, as part of the install.

Done. Simply always created on loading admin page. Formal DB update not required.

jegelstaff:
- [ ] Need to present primary relationship in privileged position in UI, and make it not editable.
- [ ] Need to make all screens default to -1 instead of 0 for relationship

jegelstaff:
Stretch goals:

- [ ] provide way of managing connections/links between forms, outside relationship UI? Or revamped relationship UI. Allow dragging forms to link them up, and prompt with wizard for connection details 

Key thing is this would allow for more obvious one to one connection. But the utility of one to one is not obvious/easily surfaced in Formulize right now, so keeping it in existing relationship UI would be OK.

modules/formulize/admin/save/relationship_settings_save.php
```diff
@@ -96,8 +96,8 @@ function addlink($form1_id, $form2_id, $relationship_id) {
 
     // write the link to the links table
     $writelink = "INSERT INTO " . $xoopsDB->prefix("formulize_framework_links") .
-        " (fl_frame_id, fl_form1_id, fl_form2_id, fl_key1, fl_key2, fl_relationship, fl_unified_display, fl_unified_delete, fl_common_value)".
```
jegelstaff, line 99: This needs reverting, we need the explicit value in the DB at all times, even though we're moving ahead with everything always being this way, and not worrying about the value, we want the data to be clean. (?) or have we/can we delete the field from the DB entirely and remove all references to it??

modules/formulize/admin/formindex.php
```diff
@@ -112,7 +111,7 @@ function patch40() {
         }
     }
 
-    if (!$needsPatch AND !file_exists(XOOPS_ROOT_PATH.'/modules/formulize/custom_code') AND (!isset($_GET['op']) OR ($_GET['op'] != 'patch40' AND $_GET['op'] != 'patchDB'))) {
+    if (!$needsPatch AND primaryRelationshipExists() AND !file_exists(XOOPS_ROOT_PATH.'/modules/formulize/custom_code') AND (!isset($_GET['op']) OR ($_GET['op'] != 'patch40' AND $_GET['op'] != 'patchDB'))) {
```
jegelstaff, line 114: We need to make the custom code detection smarter... if code folder has only one file, and there are code features to convert... needs to be its own function :(

jegelstaff: Done

PR #547, opened 2024-09-24
New formulize.org website

jegelstaff:
Revamping the site with a new theme and touchups


jegelstaff:
New site needs to include this: https://ko-fi.com/formulize

jegelstaff:
Note: to make "sections" of the website, everything keys of the path declared in the file, ie:

```
permalink: developers/API/functions/trans/
```

That's the link for the trans function documentation. There are other pages for each level of the path, ie: there's a page for developers/ and for developers/API/ etc.

So if a page is just: somehelppage/ and we want to move that under a user docs section that doesn't exist, like documentation, then the permalink needs to change to documentation/somehelppage/ and a page for the documentation section needs to be created with the path documentation/

arvinsingla:
@jegelstaff This is ready for final testing and feedback. I've incorporated the changes requested this morning.

PR #553, opened 2024-10-17
FEATURE: Configuration as code

arvinsingla:
A configuration as code implementation. Currently it supports `forms` and `elements`. The general philosophy of this feature is that the configuration file will act as the "source of truth" for what the configuration of the formulize application should be. However becuase formulize is a database driven application, in practice the feature synchronizes the system to match what's stored in the configuration. As a user makes changes to the site the deviations from the configuration file can be seen and either reverted, or a new configuration file can be exported to match what's currently in the database.

Features
- Provides an admin configuration screen which shows configuration differences between the file and the database
- Differences between the file and DB can be reverted to what is in the file
- Full configuration can be exported to a JSON file
- Detects changes to element database field types


modules/formulize/admin/formindex.php
```diff
@@ -826,7 +826,7 @@ function patch40() {
 											break;
```
arvinsingla, line 826: I genuinely don't know why these changes are showing up. I copied the file from `master` and put it in my branch to remove any conflicts yet these changes are still showing up.

modules/formulize/admin/config-sync.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
```
jegelstaff, line 2: So there's a semi standard boilerplate of comments that shows up at the top of most files. More recent ones are referencing the Formulize Project not Freeform Solutions, etc. What do you think about this in terms of the open source project? I really don't know much about the best practices here, but I feel like attaching some notice of the license and the project and copyright is a generally good idea.

jegelstaff: Most recent is I think modules/formulize/class/formulize.php which is by no means perfect, just a recent example

modules/formulize/include/formulizeConfigSync.php
```diff
@@ -0,0 +1,797 @@
...
+	 * Initialize the database connection
+	 * @return void
+	 */
+	private function initializeDatabase()
+	{
+		try {
+			// @todo we should probably be usign the $xoopsDB object
+			// But it's an older version of PDO which means that many
+			// of the newer features we're using here will need to be
+			// refactored. For now, we'll just use a new PDO object.
```
jegelstaff, line 43: Sounds good to me! There are other more esoteric parts of the admin side that don't use xoopsDB either.

modules/formulize/include/formulizeConfigSync.php
```diff
@@ -0,0 +1,797 @@
...
+			$this->db = new \PDO(
+				'mysql:host=' . XOOPS_DB_HOST . ';dbname=' . XOOPS_DB_NAME,
+				XOOPS_DB_USER,
+				XOOPS_DB_PASS
+			);
+			$this->db->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
+			$this->db->query("SET NAMES utf8mb4");
```
jegelstaff, line 50: Ah, the silly things you have to do just to get character sets to work. If you haven't suffered through character set compatibility and conversion issues, are you a real developer??

jegelstaff:
Tremendous! I haven't gone over it as closely as say the Duration element, since this is all new and you have put the main functionality through its paces. I didn't notice anything for handling absent values/defaults when values are missing from the config. I think this first pass assumes a complete set of data for each form and element? 

Also, side note, I can imagine new methods in the element classes, to handle the normalization/translation of values to/from config and the DB. ie: converting the ,1,3, group list to/from something like "Webmasters, Anonymous Users" etc. That one isn't element type specific, but we could have various helper methods all over that would encapsulate all the element logic in the right place, instead of some of the logic related to elements being here. Same thing in the validation of imported data from spreadsheets... element classes should validate their own data, instead of this huge, ugly, ancient include/import_functions.php file. But I digress.

👍

PR #558, opened 2024-11-02
Smart handling of new/updated entries when importing

jegelstaff:
Also adds support for tracking entry identity via another field, not the Entry ID.


jegelstaff:
Premature. Trying again.

PR #561, opened 2024-11-06
FEATURE: Duration Element

arvinsingla:
The duration element allows a user to enter configurable values for Days, Hours, and Minutes which result in an integer duration of minutes stored in the database. A configurable Min and Max value can also be set which includes client side and backend validation.


modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+{
+
+	var $db;
+
+	function __construct($db)
+	{
+		$this->db = &$db;
+	}
+
+	// Conversion factors to minutes
```
jegelstaff, line 32: Perhaps a style thing, but wouldn't it be more typical to have the properties defined prior to the constructor? I don't care a lot as long as this works.

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+	{
+		return new formulizeDurationElement();
+	}
+
+	// Prepare admin UI data
+	function adminPrepare($element)
+	{
+		$dataToSendToTemplate = array();
+
+		if (is_object($element) && is_subclass_of($element, 'formulizeelement')) {
```
jegelstaff, line 49: I think some things (like class names?) are actually not case sensitive in PHP? But stylistically, I believe we are using formulizeElement everywhere else?

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+			$dataToSendToTemplate['ele_value'] = $ele_value;
+		}
+
+		return $dataToSendToTemplate;
+	}
+
+	// Save admin UI data
+	function adminSave($element, $ele_value)
+	{
+		if (is_object($element) && is_subclass_of($element, 'formulizeelement')) {
```
jegelstaff, line 72: Same as above, formulizeElement?

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+			// Save min/max durations
+			$ele_value['min_minutes'] = intval($_POST['min_minutes']);
+			$ele_value['max_minutes'] = intval($_POST['max_minutes']);
+
+			// Save input box size
+			$ele_value['size'] = intval($_POST['size']);
+
+			$element->setVar('ele_value', $ele_value);
+
+			return true;
```
jegelstaff, line 87: So... the return true/false from adminSave is a bit nuanced, could perhaps be better explained in the comments... Basically, the issue comes down to the UI in the browser. The saving process is driven by asynchronous javascript requests. Page in browser doesn't change. But, if we're doing something funky here which would alter something that is in the admin UI in front of the user right now, then what they see on screen no longer matches the state in the database. If that has happened, then we should return true, because that will trigger a reload to the UI for the user, so the on screen and DB values match.

What this means in practice in this case, is that we would only return true, if the saved value of min_minutes or of max_minutes, what different from the $_POST-ed values. Same with size. Basically, if the intval resulted in a change to what the user sent, then we should update the page to match the currently saved values. If there was no change, no need to trigger a page load for the user, so we can return false.

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+				${"input_$unit"} = new XoopsFormText(
+					ucfirst($unit).":",
+					$unitMarkupName,
+					$ele_value['size'],
+					5,
+					$value,
+					false,
+					true
+				);
+				${"input_$unit"}->setExtra("min='0'");
```
jegelstaff, line 181: Not sure that '0' is necessary vs just 0 ?

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+				${"input_$unit"}->setExtra("class='formulize-duration-element-input numbers-only-textbox'");
+				${"input_$unit"}->setExtra(" onchange=\"javascript:formulizechanged=1;\" jquerytag=\"$unitMarkupName\" ");
```
jegelstaff, line 183: Interesting that this was necessary. I could have sworn there's a basic jquery bit that sets this when there's a change to any input element on screen. Huh.

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+		// Create container for inputs
+		$container = new XoopsFormElementTray($caption, ' ');
+
+		// Add input field for each enabled time unit
+		foreach ($this->timeUnits as $unit => $multiplier) {
+			$unitMarkupName = $markupName . '[' . $unit . ']';
+			if ($ele_value['show_' . $unit]) {
+				$value = isset($ele_value['values'][$unit]) ? $ele_value['values'][$unit] : '';
+				${"input_$unit"} = new XoopsFormText(
+					ucfirst($unit).":",
```
jegelstaff, line 173: This is the caption, right? This should work off the language constants, so it is translation friendly.

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+					$unitMarkupName,
+					$ele_value['size'],
+					5,
+					$value,
+					false,
+					true
+				);
+				${"input_$unit"}->setExtra("min='0'");
+				${"input_$unit"}->setExtra("class='formulize-duration-element-input numbers-only-textbox'");
```
jegelstaff, line 182: I think you can leave off the numbers-only-textbox, because the TRUE as the last param of the textbox creation above, tells the system to do it as a numbers only box? Or did you find that wasn't effective by itself?

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+					if ($amount > 0 && $ele_value['show_' . $unit]) {
+						$output .= $amount . " " . $unit . " ";
+					}
+				}
+			}
+			return new XoopsFormLabel($caption, trim($output), $markupName);
+		}
+
+		// Create container for inputs
+		$container = new XoopsFormElementTray($caption, ' ');
```
jegelstaff, line 165: Should the concatenator be a : instead of a blank space? Just a thought.

jegelstaff: Actually, I guess since the boxes are prefixed with the label of the unit, we don't want colons. Might be a nice option to add at some point, to use captions or show a X:X:X format. If you think that's easy enough / worthwhile enough, go for it now.

jegelstaff: If this is done, then it would affect the disabled display above too

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+		// Basic required validation if element is required
+		if ($element->getVar('ele_req')) {
+			$validationCode[] = "var hasValue = false;\n";
+			foreach ($this->timeUnits as $unit => $multiplier) {
+				if ($ele_value['show_' . $unit]) {
+					$validationCode[] = "if(myform['{$markupName}[{$unit}]'].value != '') hasValue = true;\n";
+				}
+			}
+			$validationCode[] = "if(!hasValue) {\n";
+			$validationCode[] = "    window.alert('Please enter a duration.');\n";
```
jegelstaff, line 214: In this and the other alerts to the user, I suggest using the caption of the element. Better user experience to direct them to where the problem is. I think we can also set focus on the element via the $markupName which is definitely a best practice too.

modules/formulize/class/durationElement.php
```diff
@@ -0,0 +1,271 @@
...
+		if ($ele_value['max_minutes'] > 0) {
+			$validationCode[] = "if(totalMinutes > {$ele_value['max_minutes']}) {\n";
+			$validationCode[] = "    window.alert('Duration must not exceed {$ele_value['max_minutes']} minutes.');\n";
+			$validationCode[] = "    return false;\n";
+			$validationCode[] = "}\n";
+		}
+
+		return $validationCode;
+	}
+
```
jegelstaff, line 245: I think you need to have the methods that are not included here, and just have them return $value or true or whatever, so that when those parts of the system kick in, you don't get a fatal error because the method is not found. I'm thinking primarily of the formatUserInputForDatabase or whatever it's called, that is used for searches, to translate user readable values into DB values. For example, if we wanted to support 1:30 as a way of representing 90 seconds, then we would have to translate that into seconds, because the search in the DB has to be done with seconds. As it stands just typing in total seconds into the search box should work (when the right method is present, even if it's doing nothing). But I wonder if we need to do more here... if the duration is 2 hours, people will want to search for 2, not 7200. (We are storing in total seconds, right?). But what search convention should this have? Maybe H:M:S ? There is a cheat sheet that tells people what all the search tricks are, this should definitely go in there. You can find the sheet by clicking the question mark on the far left of the search box row on a standard list.

jegelstaff: Sorry, I'm talking about seconds up there, and it's days, hours, minutes we're talking about. But anyway, same idea. Though days really throws a wrench in the X:X:X format. Hmmmm. This is tricky.

modules/formulize/language/english/main.php
```diff
@@ -240,6 +240,9 @@
 define("_formulize_SAVEREPORTBUTTON", "Save");
 define("_formulize_REPORTNAME", "Report Name:");
 define("_formulize_ANDORTITLE", "Interfield AND/OR Setting:");
+define("_formulize_DAYS", "Days");
```
jegelstaff, line 243: Because I'm a dick, I'm prepared to live without a complete French translation of the admin UI (mostly because it hasn't been complete for a very, very long time if ever). But the front side stuff needs a proper French translation, I consider that a minimum requirement, anything end user-facing needs English / French because we're in a Canada and if we are going to get Canadian orgs to use this, gotta be bilingual.

jegelstaff: Should just be a matter of making a main.php modification in the /french folder. But note the issue above in the render method about using these for the captions of the boxes.

jegelstaff:
So... there's at least some things here that must be addressed before merge because they will break stuff, or they are necessary for complete UX.

modules/formulize/class/durationElement.php
arvinsingla: `min` like any html attribute requires it's values to be wrapped in quotes.

modules/formulize/class/durationElement.php
arvinsingla: I could try again without, but I'm almost positive it wasn't working until I added it.

modules/formulize/class/durationElement.php
arvinsingla: After reading through the code that calls the `adminSave` method I see what you mean. It definitely wasn't clear wtihout reading that what the return true/false values were meant to signify a value change that should result in a page reload. I had assumed that it was meant as a pass/fail boolean.

modules/formulize/class/durationElement.php
arvinsingla: What might be useful as a first pass is allowing the user to enter a single value with the suffix. E.g. `1h` This will allow them to do useful searches like `> 1h` or `< 30m`

modules/formulize/class/durationElement.php
jegelstaff: Wow, did not know that. I thought it was only strings that needed them. Good help is hard to find. 😔

modules/formulize/class/durationElement.php
jegelstaff: Amazing!

PR #563, opened 2024-11-07
FEATURE: Adding a trim option to text input

arvinsingla:
This feature adds a checkbox on the options page which trims the input values before saving. This option is enabled by default for all new elements but not for existing elements.

Additionally I restructuted the options page a little for clarity. An example of the layout is below.

Screenshot 2024-11-07 at 15 21 48

### DIFF
--- a/modules/formulize/class/textElement.php
+++ b/modules/formulize/class/textElement.php
@@ -108,6 +108,7 @@ class formulizeTextElementHandler extends formulizeElementsHandler {
 				$ele_value[6] = isset($formulizeConfig['number_prefix']) ? $formulizeConfig['number_prefix'] : '';
 				$ele_value[7] = isset($formulizeConfig['number_decimalsep']) ? $formulizeConfig['number_decimalsep'] : '.';
 				$ele_value[8] = isset($formulizeConfig['number_sep']) ? $formulizeConfig['number_sep'] : ',';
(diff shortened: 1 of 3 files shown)


modules/formulize/class/textElement.php
```diff
@@ -230,6 +231,10 @@ function generateValidationCode($caption, $markupName, $element, $entry_id=false
 		// $entry_id is the ID number of the entry that this data is being saved into. Can be "new", or null in the event of a subformblank entry being saved.
     function prepareDataForSaving($value, $element, $entry_id=null) {
 			$ele_value = $element->getVar('ele_value');
+			// Trim the value if the option is set
+			if ($ele_value[12]) {
```
jegelstaff, line 235: I suggest:
```
if(isset($ele_value[12] AND $ele_value[12]) {
```
The thing with PHP is that referencing a key that does not exist in an array, results in a warning. Didn't used to, but clutters up logs now and theoretically it is a warning because there may be a logic issue in the code if non existent array keys are popping up at runtime. Not in our case, but checking with isset is still worthwhile.

jegelstaff:
Beautiful. And love the tidy up of the UI. In general, feel free to make little commits and PRs cleaning up stuff like that whenever it's tangential to whatever you might be working on.

This is all good. I made one tiny suggestion. Go for it. 👍

PR #601, opened 2024-11-27
Adding flags to relationship links to control advanced one to one behaviours

jegelstaff:


jegelstaff:
This will probably be adjusted in the Primary Relationship PR

PR #608, opened 2025-01-09
Feature: Changing color picker to html5 input color element.

arvinsingla:
Changes the colorpicker to use the `<input type="color" />` html5 element instead of the custom javascript. Since data is stored in the same format it should be a drop in replacement with no breaking changes.

- Uses input type color
- Adds option for selecting default color
- Removes color-picker.js file

 
Screenshot 2025-01-09 at 12 56 44
Screenshot 2025-01-09 at 12 56 56


modules/formulize/templates/admin/element_type_colorpick.html
```diff
@@ -0,0 +1,11 @@
+<div class="panel-content content">
+	<fieldset>
+			<legend>Default Color</legend>
+			<div class="description">
+					<p>Select the default color.</p>
```
arvinsingla, line 5: Not sure if we want something more detailed here?

jegelstaff: Or anything at all? "Default Color" kind of says it all? But should it be "Colour" ? Formulize, made in Canada.

jegelstaff:
OK, generally looks good,. Simple. I see you changed pretty much nothing in the Formulize side, it's mainly changes to how ICMS renders the 'color' form element type, that makes sense. So, questions... have you checked/verified that...

- [ ] The required behaviour works as expected when the admin makes the element required?
- [ ] The saving routine works properly when the only change in the form is to the color element? (unlike before where you had to add a textbox and change that to trigger a proper save operation)
- [ ] Can the default value be set to no color?
- [ ] What is the behaviour when the element is disabled? Do we just get the color patch, but you can't interact with it?
- [ ] Is there a way to get the HTML hex code for the selection? Or do you have to inspect/go to the database?

I think of those, only the second one is a showstopper for the Settings form colour selections. But it would be nice if the others worked as expected.

jegelstaff (commented):
Oh, I made a general comment in the conversation. This is my review comment. Submitted.

PR #611, opened 2025-01-20
Configuration as code improvements

arvinsingla:
A few improvements to configuration as code handling.

* Using array order for elements to set `ele_order` instead of explicitely setting `ele_order`
* Moved `data_type` to a `metadata` key
* Added `data_type_size` to the `metadata` key to keep correct sizing of data types

### DIFF
--- a/modules/formulize/include/formulizeConfigSync.php
+++ b/modules/formulize/include/formulizeConfigSync.php
@@ -741,8 +741,8 @@ class FormulizeConfigSync
 	private function prepareElementForExport(array $elementRow): array
 	{
 
-		$elementHandler = $this->elementHandler->get($elementRow['ele_handle']);
-		$elementDataType = $elementHandler->getDataTypeInformation();
+		$elementObject = $this->elementHandler->get($elementRow['ele_handle']);
+		$elementDataType = $elementObject->getDataTypeInformation();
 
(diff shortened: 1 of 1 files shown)


modules/formulize/include/formulizeConfigSync.php
```diff
@@ -729,8 +738,12 @@ private function exportElementsForForm(int $formId): array
 	 * @param array $elementRow Raw element data from database
 	 * @return array Prepared element data for export
 	 */
-	private function prepareElementForExport(array $elementRow, array $dataTypes): array
+	private function prepareElementForExport(array $elementRow): array
 	{
+
+		$elementHandler = $this->elementHandler->get($elementRow['ele_handle']);
```
jegelstaff, line 744: naming issue... I would strongly suggest we call this $elementObject not $elementHandler

jegelstaff (changes requested):
Just one naming issue

modules/formulize/include/formulizeConfigSync.php
jegelstaff: Committed now. Kinda wild you can invoke "Codespace" in Github and it's VSCode in a browser.

jegelstaff (approved):

PR #612, opened 2025-01-28
BUGFIX: Config-as-code decimal values weren't applied correctly

arvinsingla:
When an element type is `decimal` the result of `$elementObject->getDataTypeInformation()` will only return the decimal portion of the full configuration type (e.g. `decimal(11,2)` will return `decimal` and `2` as values. As a result when we send the dataType for applying configuration changes that included decimal values  we weren't sending the correct information.

The `getDataTypeInformation` has now been updated to include the raw string value as an additional key in the return array so we can use it directly with our configuration management.

### DIFF
--- a/modules/formulize/class/elements.php
+++ b/modules/formulize/class/elements.php
@@ -109,7 +109,7 @@ class formulizeElement extends FormulizeObject {
 			$defaultTypeSize = '';
 		}
 		//define array and return type and size
-		return array("dataType" => $defaultType, "dataTypeSize" => $defaultTypeSize, "dataTypeString" => $defaultTypeComplete);
+		return array("dataType" => $defaultType, "dataTypeSize" => $defaultTypeSize, "dataTypeCompleteString" => $defaultTypeComplete);
 
 	}
 
--- a/modules/formulize/class/elements.php
+++ b/modules/formulize/class/elements.php
@@ -109,7 +109,7 @@ class formulizeElement extends FormulizeObject {
 			$defaultTypeSize = '';
 		}
 		//define array and return type and size
-		return array("dataType" => $defaultType, "dataTypeSize" => $defaultTypeSize);
+		return array("dataType" => $defaultType, "dataTypeSize" => $defaultTypeSize, "dataTypeString" => $defaultTypeComplete);
 
 	}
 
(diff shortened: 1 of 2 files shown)


jegelstaff:
Interesting. Should this change go into getDataTypeInformation natively? Or maybe should it be another method on the element objects?

arvinsingla:
This is something I think is worth talking about. Updating getDataTypeInformation would be great but I haven't dug into where all it's being used that would require updates. In moving towards object oriented things it would be nice if there was an object for the dataType that has methods for instance to pull peices or the SQL string representation. But if we wanted to keep the format the same as it is now we could simply expand the keyed array to include additional things.

jegelstaff:
Quick check of the code tells me that the getDataTypeInformation method is referenced in a few places. But it should be easy to modify without affecting other stuff because it returns an array with multiple values labeled with keys:
```
return array("dataType" => $defaultType, "dataTypeSize" => $defaultTypeSize);
```
So it seems that if we add more metadata to what it's returning, then there you go, "dataTypeString" could be a value it returns, and that's it, done?

arvinsingla:
Yeah, that would work for my purposes. If I had the full string I could just keep that in the metadata and compare it to the result coming out the DB. It would be way easier that way.

I'll update this PR to make all those changes 👍🏽

modules/formulize/class/elements.php
```diff
@@ -109,7 +109,7 @@ function getDataTypeInformation() {
 			$defaultTypeSize = '';
 		}
 		//define array and return type and size
-		return array("dataType" => $defaultType, "dataTypeSize" => $defaultTypeSize);
+		return array("dataType" => $defaultType, "dataTypeSize" => $defaultTypeSize, "dataTypeString" => $defaultTypeComplete);
```
arvinsingla, line 112: What are your thoughts on this. Maybe `dataTypeCompleteString`?

jegelstaff: Yes, I think that would be better. There's only two hard things in computer science...

jegelstaff (changes requested):
Just the name change that you suggested

jegelstaff (approved):

jegelstaff:
Oh okay. I didn't think the review process blocked the merge. Definitely no need for that.

arvinsingla:
It's true, it didn't block the merge in this case. I just don't like merging without the Green checkmark. It's my pavolovian response 😅

PR #613, opened 2025-01-30
Adding option for how many rows to display at once in subform interfaces

jegelstaff:
- Added admin UI for number of rows (0 or blank for show all rows)
- Made subform elements work with live conditional display JS and formulize_xhr_responder
- Compartmentalized the generation of the pager UI, so subform elements and lists of entries can use it
- Trigger similar process to conditional display when user clicks a page number, to regenerate and redisplay the subform interface
- Add a limit clause to the SQL statement that handles ordering the subform entries, so that we only show the correct ones for the page the user is looking at


jegelstaff:
@arvinsingla conflicts resolved, this is ready for review. I wonder if the drawing of the subsequent pages should be faded in or animated somehow, I think it's just an instant redraw of that part of the DOM right now.

arvinsingla (approved):
LG2M Pending the discussed text changes 👍🏾

PR #616, opened 2025-01-30
Tidy up of the fde function

jegelstaff:


modules/formulize/class/applications.php
```diff
@@ -554,12 +554,12 @@ function updateMenuLink($appid,$menuitems){
         $linkValues = explode("::",$menuitems);
 	//error_log("link values ".print_r($linkValues));
         $updatesql = "UPDATE `".$xoopsDB->prefix("formulize_menu_links").
-	"` SET screen= '".formulize_db_escape($linkValues[2])."', url= '".formulize_db_escape($linkValues[3])."', link_text='".formulize_db_escape($linkValues[1])."',note='".formulize_db_escape($linkValues[6])."' where menu_id=".formulize_db_escape($linkValues[0])." AND appid=".$appid.";";
+	"` SET screen= '".formulize_db_escape($linkValues[2])."', url= '".formulize_db_escape($linkValues[3])."', link_text='".formulize_db_escape($linkValues[1])."',note='".formulize_db_escape($linkValues[6])."' where menu_id='".formulize_db_escape($linkValues[0])."' AND appid=".$appid.";";
```
jegelstaff, line 557: Interesting issue here. menu_id should be a number. What gets passed in here... can't tell from this level how reliable it is, but in practice I'm reasonably sure it's a number. Now, we could replace formulize_db_escape with intval, and ditch the quotes. We could also just keep things as is. The quotes and the escape function will work as expected. However, there is a DB layer interaction effect to take note of, which is that I'm pretty sure quotes around numeric field values are a fatal error in normal MySQL/MariaDB queries. The old queries throughout the system do this kind of thing a lot though. And they do things like include the PK in an insert query, but have the value empty, and rely on the DB to fill it in with the autoincrement, which is also bad form and a fatal error since version I don't know what.

To get around the compatibility issues there, we explicitly turn off "strict mode" in the database, when making the DB connection. Ever since strict mode was on by default, that was necessary to prevent the system from being completely unusable. And so, with strict mode blissfully off, bad things like quotes around numeric values are allowed to continue. But I kind of think this should switch to intval.

modules/formulize/class/applications.php
```diff
@@ -554,12 +554,12 @@ function updateMenuLink($appid,$menuitems){
...
         if(!$result = $xoopsDB->query($updatesql)) {
             exit("Error updating Menu Item. SQL dump:\n" . $updatesql . "\n".$xoopsDB->error()."\nPlease contact <a href=mailto:info@formulize.org>info@formulize.org</a> for assistance.");
         }else{
         	//delete existing permissions for this menu item
-        	$deletepermissions = "DELETE FROM `".$xoopsDB->prefix("formulize_menu_permissions")."` WHERE menu_id=".formulize_db_escape($linkValues[0]).";";
+        	$deletepermissions = "DELETE FROM `".$xoopsDB->prefix("formulize_menu_permissions")."` WHERE menu_id='".formulize_db_escape($linkValues[0])."';";
```
jegelstaff, line 562: Same here. intval.

modules/formulize/class/forms.php
```diff
@@ -1324,10 +1324,10 @@ function buildPerGroupFilterWhereClause($match,$indexes,$filterSettings,$uid,$fo
                     if($filterSettings[1][$i] == "NOT") {
                         $filterSettings[1][$i] = "NOT LIKE";
                     }
-                    $perGroupFilter .= "($formAlias`".$filterSettings[0][$i]."` ".htmlspecialchars_decode($filterSettings[1][$i]) . " " . "\"$likeBits"."*=+*:".formulize_db_escape($termToUse)."*=+*:"."$likeBits\" OR $formAlias`".$filterSettings[0][$i]."` ".htmlspecialchars_decode($filterSettings[1][$i]) . " " . "\"$likeBits"."*=+*:".formulize_db_escape($termToUse)."\")";
+                    $perGroupFilter .= "($formAlias`".$filterSettings[0][$i]."` ".htmlspecialchars_decode($filterSettings[1][$i]) . " " . "\"$likeBits"."*=+*:".formulize_db_escape($termToUse)."*=+*:"."$likeBits\" OR $formAlias`".$filterSettings[0][$i]."` ".htmlspecialchars_decode($filterSettings[1][$i]) . " " . "'$likeBits"."*=+*:".formulize_db_escape($termToUse)."')";
```
jegelstaff, line 1327: First instance of likeBits needs \" converted to '

modules/formulize/class/data.php
```diff
@@ -607,7 +607,7 @@ function findAllEntriesWithValue($element_id, $value, $scope_uids=array(), $scop
 		global $xoopsDB;
     $form_handler = xoops_getmodulehandler('forms', 'formulize');
     $formObject = $form_handler->get($this->fid);
-		$queryValue = "\"" . formulize_db_escape($value) . "\"";
+		$queryValue = "'" . formulize_db_escape($value) . "'";
```
jegelstaff, line 610: We need to test things going in and out of the db with this change. This is fundamental. And what happens to single and double quotes now that we're single quoting... ack. Prediction, they're all special characters. But when they're not special characters, this will still work right??? Need to verify now.

jegelstaff:
Need to verify that character set is set with DB connection somehow. Quote method depends on it.

modules/formulize/class/templateScreen.php
```diff
@@ -84,10 +84,10 @@ function insert($screen) {
         $screen->assignVar('sid', $sid);
 
         if (!$update) {
-            $sql = sprintf("INSERT INTO %s (sid, custom_code, template, donedest, savebuttontext, donebuttontext, viewentryscreen) VALUES (%u, %s, %s, %s, %s, %s, %s)", $this->db->prefix('formulize_screen_template'),
+            $sql = sprintf("INSERT INTO %s (sid, custom_code, template, donedest, savebuttontext, donebuttontext, viewentryscreen) VALUES (%u, '%s', '%s', '%s', '%s', '%s', '%s')", $this->db->prefix('formulize_screen_template'),
```
jegelstaff, line 87: This one is ok because the value being passed in is being put through quoteString

modules/formulize/class/templateScreen.php
```diff
@@ -84,10 +84,10 @@ function insert($screen) {
...
                 $screen->getVar('sid'), $this->db->quoteString(formulize_db_escape($screen->getVar('custom_code'))), $this->db->quoteString(formulize_db_escape($screen->getVar('template'))), $this->db->quoteString(formulize_db_escape($screen->getVar('donedest'))), $this->db->quoteString(formulize_db_escape($screen->getVar('savebuttontext'))), $this->db->quoteString(formulize_db_escape($screen->getVar('donebuttontext'))), $this->db->quoteString(formulize_db_escape($screen->getVar('viewentryscreen'))));
         } else {
-            $sql = sprintf("UPDATE %s SET custom_code = %s, template = %s, donedest = %s, savebuttontext = %s, donebuttontext = %s, viewentryscreen = %s WHERE sid = %u", $this->db->prefix('formulize_screen_template'),
+            $sql = sprintf("UPDATE %s SET custom_code = '%s', template = '%s', donedest = '%s', savebuttontext = '%s', donebuttontext = '%s', viewentryscreen = '%s' WHERE sid = %u", $this->db->prefix('formulize_screen_template'),
```
jegelstaff, line 90: Same here

jegelstaff:
> Need to verify that character set is set with DB connection somehow. Quote method depends on it.

Done. Presently utf8 as defined in the DSN in the mainfile.php. New installs will be utf8mb4 as of ab0bca56b24d77895aa9a891ac15be4afe3ab8bf and the DSN could be changed in existing installs to ensure full 4 byte unicode support. DB collations of tables and fields may need to be updated in older systems for full compatibility. But bottom line is that unicode is the character set, we're not going to have some alternate character set messing up how the quoting works.

PR #627, opened 2025-03-04
Refactor display function, change title > handle in the way datasets are built

jegelstaff:
* Stop using the form title for the key identifying the form in the getData results, use form_handle instead

### DIFF
--- a/modules/formulize/include/extract.php
+++ b/modules/formulize/include/extract.php
@@ -1243,9 +1243,9 @@ function processGetDataResults($resultData)
 					// will be collected with mainform ($fid) data
 					if ($field == 'main_email') {
 						if ($is_webmaster or $view_private_fields or $creatorAllowsEmailViewing or $creatorUid == $this_userid) {
-							$masterResults[$masterIndexer][getFormTitle($fid)][$entryIdIndex['main']]['creator_email'] = $value;
+							$masterResults[$masterIndexer][getFormHandle($fid)][$entryIdIndex['main']]['creator_email'] = $value;
 						} else {
-							$masterResults[$masterIndexer][getFormTitle($fid)][$entryIdIndex['main']]['creator_email'] = "";
+							$masterResults[$masterIndexer][getFormHandle($fid)][$entryIdIndex['main']]['creator_email'] = "";
 						}
 					}
 					continue;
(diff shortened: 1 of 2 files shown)


jegelstaff:
@tikyon Hi Parlin, I believe this branch fixes an issue in 7.41 that doesn't come up all the time, but is a fatal error when it does come up. In addition, I went and changed a bunch of old stuff where the form title was being used where really the form handle should have been used (but form handles didn't exist when the original code was written). I believe the change is complete and won't have bad effects on existing behaviours. But I would really appreciate it if you could checkout this branch (FIX/scalar-global-error-in-display) and try out any test site you might have available with these changes.

I think any issues would be fairly glaring, since the big change is to how the main array returned by getData is constructed, it now uses form handles instead of the form title for the key that identifies the forms. So any code that is referencing that by form title explicitly would fail. Such code would need to be updated to use form handles. But I think I've accounted for the parts of the code that actually care. Still, some additional verification before it gets merged in would be greatly appreciated!

Thanks!

jegelstaff:
Update: these two points were pulled out of this and put in master now:
* Stop putting the value from the DB into a single item array, just to pull it out again in the display function
* Fix the interaction with $GLOBALS that could lead to fatal error in some circumstances

jegelstaff:
Only issue in here is the refactoring to actually use form handle instead of form title as the key in the array. Less pressing issue, still may have consequences that need evaluation.

tikyon:
The first issue I run into is that it broke fetching of linked forms when using `internalRecordIds($entry, $fid)` returning empty arrays. The data for the linked forms are available in `$entry` array.

jegelstaff:
Ack! I had a thought that internalRecordIds might be affected, and I forgot to look into it. Thanks for the report. I'll make a commit and let you know when to check again.

jegelstaff:
@tikyon I believe the latest commit to this branch fixes the internalRecordIds problem. And a review of code makes be cautiously optimistic there are no other references to the form title where we would actually want (and use) form handle now.

jegelstaff:
Note that if you have code that is manually fiddling with the getData array, it may be expecting form titles at a certain point, and they will be handles now, so that code would need updating to be compatible with this. But this is very much a forward looking change and worthwhile because using titles in the getData array results was always always a dumb idea. But no form handles way back when, and it was always a sort of internal issue, so it was left alone. But these internal odd things are a barrier to developers and we want to remove any and all barriers as much as possible.

tikyon:
Tested 7183a6e77 and internalRecordIds issue I had earlier is fixed.

tikyon (approved):

jegelstaff:
@tikyon A further commit went in (b0fa44e058b99c6859b9ec89c950ccad86babb2f) to address an issue with derived values arising from this PR. Very important to use the latest rather than yesterday's Github version.

Issue #626, opened 2025-03-03
Recurring PHP warnings in v7.41


tikyon:
- Undefined global variable $formulize_customCodeForApplications
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/customCodeForApplications.php#L16

- Undefined global variable $formulize_checkForLinks_oneToOneMetaData
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/readelements.php#L222 
 
- Together: Undefined array key 3 / Undefined array key 4 / Undefined array key 5
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L544

- Undefined array key "main", line 803
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L803

- Undefined variable $existsJoinText
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L714
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L812

- Undefined array key 1
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L1507

- Undefined array key 2
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L1517

- Together: Undefined array key {number} / Undefined array key "and"
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L1861

- Undefined global variable $formulize_doingExport / $formulize_returnAfterSettingBaseQuery
https://github.com/jegelstaff/formulize/blob/fe5f77b51c7dc9a39fd7b6161a0fa435f37eaabd/modules/formulize/include/extract.php#L167
[...]

PR #628, opened 2025-03-04
Fixing various notices and warnings

jegelstaff:


jegelstaff:
See issue #626

modules/formulize/include/extract.php
```diff
@@ -1501,8 +1501,13 @@ function formulize_parseFilter($filtertemp, $andor, $linkfids, $fid, $frid)
 			// Use the local andor setting ($filterParts[0]) between terms
 
 			$ifParts = explode("/**/", $indivFilter);
-
-			// FINAL NOTE ABOUT SLASHES...Oct 19 2006...patch 22 corrects this slash/magic quote mess.  However, to ensure compatibility with existing Pageworks applications, we are continuing to strip out all slashes in the filterparts[1], the filter strings that are passed in, and then we apply HTML special chars to the filter so that it can match up with the contents of the DB.  Only challenge is that extract.php is meant to be standalone, but we have to refer to the text sanitizer class in XOOPS in order to do the HTML special chars thing correctly.
+			// if the indivFilter does not have all possible values, provide empty strings for the missing ones
+			if(!isset($ifParts[1])) {
+				$ifParts[1] = "";
+			}
+			if(!isset($ifParts[2])) {
```
tikyon, line 1508: This if-condition broke textbox search when not using any operator.
> Error: could not count master results.
> You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ''Test') 

The problem part looks something like `source.handle_name 'Test'` with "Test" being the search term used in the textbox.

tikyon: When it got to line 1529 `$operator = isset($ifParts[2]) ? $ifParts[2] : "LIKE";`, it assigned the empty string that was initialized above.

jegelstaff:
If I understand correctly, changing the new code to assign LIKE as default when !isset instead of "" will solve the problem?

modules/formulize/include/extract.php
jegelstaff: If I understand correctly, changing the new code to assign LIKE as default when !isset instead of "" will solve the problem? If so, go ahead and commit that to this branch.

tikyon (approved):
Looks good, and resolved the first 7 bullets in #626

jegelstaff:
@tikyon More changes, I believe everything in the referenced issue is now addressed. Do you recall/have clear test cases for all of them?? :) Let me know what you think of these.

tikyon (approved):
Thanks, Julian. These look good to me.

PR #633, opened 2025-03-09
Pref for Revisions History always on

jegelstaff:
Added a module preference in Formulize for whether revisions history is always on for all form. If so, setting on admin side is disabled and always on.

If preference is on, form objects always have store_revisions set to 1.

Interactions with revisions tables are always wrapped in checks for if the revision table exists. If the preference is on, then the revision table is created as part of this check, if it doesn't already exist.


jegelstaff:
Replaced by #634

PR #637, opened 2025-03-13
Delete entries if not present in import

jegelstaff:
Carefully limit deletion to only entries the user could normally delete. Also allow them to limit the deletion, in case they are uploading only a subset of entries, and just want to delete others not present in the sheet only if they are part of the subset. ie: uploading June block students, so delete students not in sheet, but only if they're June block in the DB, because the upload is taken as authoritative.


jegelstaff:
- [ ] Option should only be visible if user has delete perm of some kind

PR #644, opened 2025-03-26
FEATURE: Adding support for database options in trust config

arvinsingla:
Provides an SDATA_DB_OPTIONS constant in your trust file which can be used to add PDO options. If the constant is missing or is an empty array the default of `null` will be passed to PDO.

### DIFF
--- a/install/templates/sdata.dist.php
+++ b/install/templates/sdata.dist.php
@@ -39,3 +39,8 @@ define('SDATA_DB_PREFIX', '');
 // This salt will be appended to passwords in the icms_encryptPass() function.
 // Do NOT change this once your site is Live, doing so will invalidate everyones Password.
 define('SDATA_DB_SALT', '');
+
+// Additional database options
+// If you need to specify additional options for the database connection, you can do so here.
+// For example, to specify the charset to be used, you can set the PDO::MYSQL_ATTR_INIT_COMMAND option to "SET NAMES utf8"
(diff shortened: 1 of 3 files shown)


install/templates/sdata.dist.php
```diff
@@ -1,41 +1,46 @@
...
+// Table Prefix
+// This prefix will be added to all new tables created to avoid name conflict in the database. If you are unsure, just use the default 'icms'
+define('SDATA_DB_PREFIX', '');
+
+// Password Salt Key $mainSalt
+// This salt will be appended to passwords in the icms_encryptPass() function.
+// Do NOT change this once your site is Live, doing so will invalidate everyones Password.
+define('SDATA_DB_SALT', '');
+
+// Additional database options
```
jegelstaff, line 43: Very pedantic point... The charset may not be the best example because the charset is controlled in another constant that mainfile.php references, and it is included directly in the dsn. Referencing the certificate might be best. But kind of doesn't matter.

jegelstaff (approved):
One totally minor comment. Depending how you feel about it, merge away.

install/templates/sdata.dist.php
arvinsingla: Yeah no worries, I will update the example. Maybe setting the ATTR_ERRMODE would be a good example since it's a single item.

PR #645, opened 2025-03-26
FIX: onetoone and regular conditions work together on the same form

jegelstaff:
This PR cleans up the handling of conditional elements, when there is a mix of onetoone elements and regular conditional elements on the same form screen.

Previously, multiple conditional checks were being fired, when we only want the one to one conditions in cases where the one to one governing element was altered.

There are various fixes/refinements to determining the correct entry to use as the context for rendering the element, in various situations.

One to one conditions now have the one to one metadata attached, and we respond to that accordingly. Non-onetoone conditions don't have that metadata and behave normally.

### DIFF
--- a/modules/formulize/formulize_xhr_responder.php
+++ b/modules/formulize/formulize_xhr_responder.php
@@ -181,6 +181,7 @@ switch($op) {
     include_once XOOPS_ROOT_PATH . "/modules/formulize/include/extract.php";
     $sendBackValue = array();
     $element_handler = xoops_getmodulehandler('elements','formulize');
+    $elementsToProcess = array();
     foreach($_GET as $k=>$v) {
       if($k == 'elementId' OR $k == 'entryId' OR $k == 'fid' OR $k == 'frid' OR substr($k, 0, 8) == 'onetoone' OR $k == 'sid') { // serveral onetoone keys can be passed back too
         if($k == 'onetooneentries' OR $k == 'onetoonefids') {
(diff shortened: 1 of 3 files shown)


jegelstaff:
Still some work to do here perhaps, in making sure that onetoone flag is passed to checkConditions in the .js file, properly for the relevant element set that is being passed in?? Right now, onetoone flag will be set based on the last element processed, which seems a bit wrong??

jegelstaff:
@arvinsingla This is ready I think. This resolves a very esoteric issue that appears in Clinics. Ironically, I think we should modify the form UI in Clinics in a way that would sidestep this issue (see below). None the less, this is an important change for compatibility with a variety of situations that could crop up in real life in other scenarios.

The issue, as described above, appeared in Clinics, when the conditional element "Age This Year" (dependent on the year of birth not being blank) would not refresh itself when the Participants form was displayed in a one-to-one relationship with the Participants in Clinics form. This is found in the Clinics site by going to the details page for a Clinic, and then clicking on the edit icon for one of the participants in the clinic, and what the user lands on is the Participants in Clinics form, plus the Participants form (that shows the year of birth, phone number, etc).

The problem was that the conditional logic was tripping over itself in a few places and was not distinguishing between conditional triggers that involved the one-to-one connection, vs regular ones (like the age/dob). The one-to-one connection triggers conditional behaviour if you change the Participant. What happens in that case, if you switch from David Bowie to Tom Jones, for example, is that all the elements in the Participants form switch over to show the new person's data (phone number, dob, etc).

With this PR, that behaviour works cleanly in both cases, on that subform (the regular condition on the Participants form, without the one to one, always worked fine even prior to this PR).

So, I would like your review of this, because this is important functionality that should be applied to master.

And it should be applied to Clinics, to resolve the issue. HOWEVER, as mentioned above, I think we can sidestep the issue there. Really, we should not allow the Participants element to be editable in the one-to-one form screen (Participants in Clinics form plus Participants form), one it has a value. I think we make it disabled for all users, but put a disabled condition on that so that it is only disabled when the value of itself is NOT {BLANK}. That way, people can fill it in to add a participant to a clinic, but then once they've done that, they won't be changing the person, and indeed it kind of doesn't make any sense to let them change the person. They're not going to turn Participant X into Participant Y. They're going to delete Participant X if they're not participating anymore. It's radically confusing to allow them to change the participant. All that disabling stuff, including the condition, is done on the Display tab of the Elements settings.

Nonetheless, even if the element is disabled, I think we probably still need this PR merged in to fix the issue, because I suspect that without this PR, even with a disabled Participant element, the Year of Birth/Age this year condition will fail, due to the presence of the one to one condition on the page. It's possible that disabling the key element that joins the forms together, will remove the one to one condition from consideration. But I suspect that's not the case and so we need this PR to smarten up the conditional logic so the regular condition won't get tripped up by the presence of the one to one connection.

jegelstaff:
@arvinsingla This is rebased on master now and ready for review.

jegelstaff:
@arvinsingla Re-rebased on master again. Really, I'm most concerned with you reading the JS code, a whole setup and review of everything like with the subform rows is not totally required, it's a very arcane setup (the one-to-one connection in Clinics for Participants and Participants-in-Clinics being the example). But I want to deploy this out to Clinics soon, it should go in before they go live.

jegelstaff:
@arvinsingla I'm going to merge this to get it in and off the plate. Your eyes on the code would still be appreciated and if there's something serious we'll make a further PR.

PR #646, opened 2025-03-28
FIX: Fixing live derived values not working properly

arvinsingla:
The quoting of the globals term was incorrect and causing issues when iterating through the formula.

Currently only tested on placements attempting to reproduce reported issue. Works as expected

### DIFF
--- a/modules/formulize/include/extract.php
+++ b/modules/formulize/include/extract.php
@@ -2208,7 +2208,7 @@ function formulize_includeDerivedValueFormulas($metadata, $formHandle, $frid, $f
 					} else {
 						$replacement = "display(\$entry, '$newterm')";
 					}
-					$replacement = "(isset(\$GLOBALS['formulize_asynchronousFormDataInAPIFormat'][\$entry_id]['$newterm']) ? \$GLOBALS['formulize_asynchronousFormDataInAPIFormat'][\$entry_id]['$newterm\'] : $replacement)";
+					$replacement = "(isset(\$GLOBALS['formulize_asynchronousFormDataInAPIFormat'][\$entry_id]['$newterm']) ? \$GLOBALS['formulize_asynchronousFormDataInAPIFormat'][\$entry_id]['$newterm'] : $replacement)";
 					$quotePos = $quotePos + strlen($replacement);
 					$formula = str_replace($term, $replacement, $formula);
 				} else {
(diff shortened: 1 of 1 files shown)


jegelstaff (commented):
Fascinating. It's not clear to me why this leads to a different result, since the outer double quotes around the string should result in $newterm being evaluated when building the string, in both cases. Whether we have single quotes or escapes double quotes on either side shouldn't matter?

It's also not clear to me how this would have resulted in the value of $newterm being used as the value of the handle in the derived value, when it would seem like the value in the GLOBALS array that has the key of the current entry id and then the $newterm would be used? Is the value of that item in the array being set wrong in formulize_xhr_responder?

And yet, if this change fixes the issue in placements, then it fixes the issue... so??

jegelstaff:
A good idea, but I don't think that's the reason, because the quotePos is set to the length of $replacement, and the $replacement string is the entire expression, including the double quoted key names. So it should start looking at the very end. If it were finding the wrong double quote, that would create an infinite loop, but the page is rendering. So there's more to this. I wonder if it's something deep in PHP about referencing a key in GLOBALS that doesn't exist and you get back the value of the key or something??

arvinsingla:
> A good idea, but I don't think that's the reason, because the quotePos is set to the length of $replacement, and the $replacement string is the entire expression, including the double quoted key names. So it should start looking at the very end. If it were finding the wrong double quote, that would create an infinite loop, but the page is rendering. So there's more to this. I wonder if it's something deep in PHP about referencing a key in GLOBALS that doesn't exist and you get back the value of the key or something??

Looking at it again, I think you're incorrect. $replacement is the string we are passing into str_replace so if there are multiple instances being replaced it str_pos is only going to increment the length of one instance.

jegelstaff:
Yes, confirmed, I checked this out locally. Yes, it is related to the structure of the formula for the name, which has multiple instances of the references to certain element handles. And because the str_replace replaces all of them, not just the one that was found first, then having double quotes in the replacements further down the line, means they will then be discovered and replaced, and so on, and then eventually the quotePos is greater than the length of the string, except the last instances of handles haven't even been replaced yet! So that's why the literal student_name etc was showing up, because it was never being replaced. So I am 100% satisfied that this fixes the issue and addresses the entire issue, because all of the behaviour is explained by this problem, and is fixed with this change.

jegelstaff:
I say go ahead and merge

PR #647, opened 2025-03-28
FEATURE: Adding http security headers

arvinsingla:
First pass at adding security headers.

Any third party data that is fetched externally needs to be added to the CSP headers.

To prevent the need for modifying core files in order to add more Content-Security-Policy headers we should also add a site configuration option to allow for the customization of this header.


docker/mariadb/conf.d/lower_case_table_names.cnf.macos
```diff
@@ -0,0 +1,2 @@
+[mariadb]
```
jegelstaff, line 1: Another pedantic comment... Would be nice to alter the readme or whatever the file is in the Docker/MariaDB folder or something like that, where it mentions the windows version of this, to just mention the macos one too.

index.php
```diff
@@ -24,6 +24,26 @@
...
+/**
+ * HTTP Security Headers
+ */
+// HTTP Strict Transport Security (HSTS) - two years
+header("Strict-Transport-Security: max-age=63072000");
+// Content Security Policy (CSP) - starting with recommended locked down policy
+header("Content-Security-Policy: default-src 'none';
+	img-src 'self';
+	script-src 'self' ajax.googleapis.com use.fontawesome.com;
+	style-src 'self' fonts.googleapis.com ajax.googleapis.com use.fontawesome.com;
```
jegelstaff, line 36: Best way to figure out if we've got everything we need, is to look at page source of a standard page? Just wondering about how to make sure we're not missing something that would break stuff. This list looks right to me off the top of my head, but would be nice if we can verify somehow.

jegelstaff (approved):

jegelstaff:
@copilot Can you review this an offer an opinion of whether it's ready for production? We're concerned about it breaking existing behaviours.

docker/docker-db-help.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Setting up a database
+SQL files in the `/docker/mariadb/seed/` folder will be parsed and run (in alphabetical order) when the MariaDB Docker container starts, if there is no database already in the `/docker/mariadb/data/` folder.
+
+Note: make sure your sql files are UTF8 encoded!! Depending on OS, etc, you may need to convert the encoding of the file prior to starting Docker.
+
+## Persisting a database between sessions
+The MariaDB database files will be stored in the `/docker/mariadb/data/` folder and any files there will be used by MariaDB when the container is next instantiated.
```
arvinsingla: This help file is outdated since we switched over to volumes for storing the database because of the cross file system issues. Which means the text above for macos and windows is also no longer necessary.

docker/mariadb/conf.d/lower_case_table_names.cnf.macos
arvinsingla: This file is no longer necessary.

jegelstaff:
@arvinsingla Claude points out that since all requests for the real application route through modules/formulize/index.php the changes in the main index.php have extremely narrow effects. And there are wideranging implications of applying what is in index.php more broadly across the whole site. Lots of things will likely break. We never noticed in the one site where this is happening, because it's applied so narrowly. A plan file is being committed to outline what can/should be done with this.

jegelstaff:
@arvinsingla Not saying this is work for you to do! Just FYI.

PR #648, opened 2025-04-01
FIX: Fix issue with DB options for existing formulize installations

arvinsingla:
Ensure that users with existing installations who don't update their mainfile.php don't get php exceptions.


jegelstaff (approved):
Excellent

PR #652, opened 2025-04-29
Fix "random" session loss

jegelstaff:
Previously, we were storing a bunch of things in the user session, that we didn't need to. If too much stuff built up in the session, which is stored in the database, then the session could no longer be decoded by PHP after retrieving the session record from the database.

To resolve this, we are no longer storing stuff in the session that we don't need to have there. The main thing is the column settings for every rendered element, which was being stored there just in case the element was conditional and we needed to render it some time in the future. This was originally thought necessary because we don't have the screen settings at the time the conditional elements are rendered, so we cached the column settings for all elements when the screen was first rendered, so when asynch requests came in later we would know the column settings to use. Now we pass the screen id with the asynch request and we use that to determine the column settings based on the screen. Duh.

We also used the old token system to generate unique, personalized tokens related to entry locking. We then stored data about which entries were locked, in a file based on the token name, so we could look up the data later when a "delete entry lock" request came in. The old token system stores every token it creates in the session, because it presumes you are validating against the session in the future. We validate entry locks against the files we write (we used to do it all in session, I think, which is maybe why it made sense originally -- but as soon as we weren't validating against data in session, then we don't need to store tokens in session any longer). So now we just generate a unique token per page load that is used to handle entry locks, and nothing is stored in the session related to this at all. Files written to disk related to entry locking, are unaffected by this change and continue to work normally.

End result is that sessions do not have unnecessary things filling them up which can lead to the session being lost when it is no longer readable. The core tokens that are used for their original anti CSRF purpose, are still stored in session. We could/should evolve this to get rid of them sooner, or store them elsewhere, due to the 64KB size limit on data stored in TEXT fields in the database, because if someone exceeded that size on a particularly long session with intensive form loading, they would lose their session still. This is only an issue if users load many tabs/windows, and do not save data in the form. Tokens are issued for every form, and cleared when validated, but tokens that are never validated hang around because many that form will be submitted later.

*I have tested this locally and entry locking still works, and columns are used correctly in asynch rendering.*

### DIFF
--- a/modules/formulize/class/fileUploadElement.php
+++ b/modules/formulize/class/fileUploadElement.php
@@ -117,9 +117,10 @@ class formulizeFileUploadElementHandler extends formulizeElementsHandler {
     // $caption is the prepared caption for the element
     // $markupName is what we have to call the rendered element in HTML
     // $isDisabled flags whether the element is disabled or not so we know how to render it
-    // $element is the element object
+		// $element is the element object
     // $entry_id is the ID number of the entry where this particular element comes from
-    function render($ele_value, $caption, $markupName, $isDisabled, $element, $entry_id) {
+    // $screen is the screen object that is in effect, if any (may be null)
+    function render($ele_value, $caption, $markupName, $isDisabled, $element, $entry_id, $screen, $owner) {
         // ele_value[3] is the fileName or error message
         // ele_value[4] is the displayName version of the fileName (only going to be diff from fileName if there is no direct linking to files allowed)
         // ele_value[5] is the flag for whether there's a file or not
--- a/modules/formulize/formulize_xhr_responder.php
+++ b/modules/formulize/formulize_xhr_responder.php
@@ -182,7 +182,7 @@ switch($op) {
     $sendBackValue = array();
     $element_handler = xoops_getmodulehandler('elements','formulize');
     foreach($_GET as $k=>$v) {
-      if($k == 'elementId' OR $k == 'entryId' OR $k == 'fid' OR $k == 'frid' OR substr($k, 0, 8) == 'onetoone') { // serveral onetoone keys can be passed back too
+      if($k == 'elementId' OR $k == 'entryId' OR $k == 'fid' OR $k == 'frid' OR substr($k, 0, 8) == 'onetoone' OR $k == 'sid') { // serveral onetoone keys can be passed back too
         if($k == 'onetooneentries' OR $k == 'onetoonefids') {
             ${$k} = unserialize($v);
         } else {
@@ -208,6 +208,14 @@ switch($op) {
         $GLOBALS['formulize_asynchronousFormDataInAPIFormat'][$passedEntryId][$handle] = $apiFormatValue;
       }
     }
+		$screenObject = null;
+		if($sid) {
+			$screen_handler = xoops_getmodulehandler('screen', 'formulize');
+			if($screenObject = $screen_handler->get($sid)) {
+				$screen_handler = xoops_getmodulehandler($screenObject->getVar('type').'Screen', 'formulize');
+				$screenObject = $screen_handler->get($sid);
+			}
+		}
 		// Normally, the entryId we're rendering is the one displayed in the form at load time, elements are dependent on conditions, but always rendered as in that entry.
 		// In a one-to-one situation, if the relationship is based on a linked element, we need to render elements from the entry selected in the governing element
 		// If the relationship is common value then we need to try to determine which entry is connected to it, if any
@@ -238,7 +246,7 @@ switch($op) {
 			$html = "";
       $json .= $jsonSep.'{ "handle" : '.json_encode('de_'.$_GET['fid'].'_'.$_GET['entryId'].'_'.$thisElementId);
       if(security_check($fid, $entryId)) {
-        $html = renderElement($elementObject, $entryId);
+        $html = renderElement($elementObject, $entryId, $frid, $screenObject);
         $json .= ', "data" : '.json_encode($html);
       } else {
        	$json .= ', "data" : '.json_encode('{NOCHANGE}');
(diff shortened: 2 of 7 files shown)


modules/formulize/class/fileUploadElement.php
```diff
@@ -117,9 +117,10 @@ function loadValue($value, $ele_value, $element) {
     // $caption is the prepared caption for the element
     // $markupName is what we have to call the rendered element in HTML
     // $isDisabled flags whether the element is disabled or not so we know how to render it
-    // $element is the element object
+		// $element is the element object
     // $entry_id is the ID number of the entry where this particular element comes from
-    function render($ele_value, $caption, $markupName, $isDisabled, $element, $entry_id) {
```
jegelstaff, line 122: This and related function signature changes are simply bringing some element class files up to date with the current standard. We do pass the screen to the render method, though it is ignored in all standard elements, so I think that is only used in some custom element for ESAT or someone else.

modules/formulize/formulize_xhr_responder.php
```diff
@@ -329,10 +337,10 @@
 
 }
 
-function renderElement($elementObject, $entryId) {
+function renderElement($elementObject, $entryId, $frid, $screenObject) {
 
 	$GLOBALS['formulize_asynchronousRendering'][$elementObject->getVar('ele_handle')] = true;
-	$deReturnValue = displayElement("", $elementObject, $entryId, false, null, null, false); // false, null, null, false means it's not a noSave element, no screen, no prevEntry data passed in, and do not render the element on screen
+	$deReturnValue = displayElement("", $elementObject, $entryId, false, $screenObject, null, false); // false, null, null, false means it's not a noSave element, no screen, no prevEntry data passed in, and do not render the element on screen
```
jegelstaff, line 343: This is just to play nice, so the render method of the element will receive the screen object, but as mentioned above, that is not actually relevant in any standard elements

modules/formulize/include/formdisplay.php
```diff
@@ -118,15 +118,8 @@ public function getTemplate($type) {
         $template = '';
 				global $xoopsUser;
 				$uid = $xoopsUser ? $xoopsUser->getVar('uid') : 0;
-        if($this->getTitle() != 'formulizeAsynchElementRender' AND is_object($this->screen)) {
```
jegelstaff, line 121: Don't need to do any session checking stuff anymore!

modules/formulize/include/formdisplay.php
```diff
@@ -241,85 +234,35 @@ public function renderValidationJS($withtags = true) {
 		return $js;
 	}
 
-    // reset will cause the cached copy of columns to be bypassed - this is done when a form is rendered, so that we don't reuse the setting for an element when it appeared on a different screen in the past in this same session
```
jegelstaff, line 244: Main culprit where we were storing and referencing stuff in session related to columns, now we can go off the screen object all the time

modules/formulize/include/formdisplay.php
```diff
@@ -241,85 +234,35 @@ public function renderValidationJS($withtags = true) {
...
+					$columns = array(1, 'auto', 'auto');
+			}
+			return $columns;
     }
 
 	function _drawElements($elements, $ret, $hidden) {
 
 		foreach ( $elements as $ele ) {
 
-            if(is_string($ele) AND isset($GLOBALS['formulize_renderedElementHasConditions'])) {
```
jegelstaff, line 297: Awkward stuff that was being used to call _getColumns to basically cache the data in the session for all elements even if not rendered this page load, so we were ready for conditional rendering. Yucky. All gone now!

modules/formulize/include/functions.php
```diff
@@ -6319,7 +6319,7 @@ function formulize_xhr_send(op,params) {
 // provides the token that is associated witht he entry locking/unlocking for this page load
 function getEntryLockSecurityToken() {
     static $token;
-    $token = $token ? $token : $GLOBALS['xoopsSecurity']->createToken(0, 'formulize_entry_lock_token');
+    $token = $token ? $token : hash('sha256',(uniqid(rand(), true).$_SERVER['HTTP_USER_AGENT'].XOOPS_DB_PREFIX.$_SERVER['REMOTE_ADDR']));
```
jegelstaff, line 6322: No need to use session token system to generate unique tokens for entry locking!

modules/formulize/include/js/conditional.js
```diff
@@ -114,7 +114,7 @@ function checkCondition(relevantElementSet, elementValuesForURL) {
         }
         elementIdsSep = ',';
     }
-	return jQuery.post(FORMULIZE.XOOPS_URL+"/modules/formulize/formulize_xhr_responder.php?uid="+FORMULIZE.XOOPS_UID+"&op=get_element_row_html&elementId="+elementIds+"&entryId="+entryId+"&fid="+fid+elementValuesForURL);
+	return jQuery.post(FORMULIZE.XOOPS_URL+"/modules/formulize/formulize_xhr_responder.php?uid="+FORMULIZE.XOOPS_UID+"&sid="+FORMULIZE.SCREEN_ID+"&op=get_element_row_html&elementId="+elementIds+"&entryId="+entryId+"&fid="+fid+elementValuesForURL);
```
jegelstaff, line 117: The critical passing of screen id to the conditional rendering request. Note that there is security about whether the user has permission to view the element, otherwise we don't render anything. No security specifically on the validity of the screen id, however screens don't contain any data related to permission-controlled activities, and all such activities (like writing data, etc) validate the user prior to doing the action. So I think we're in the clear for accepting the screen id as is.

arvinsingla (approved):
If we wanted to tighten up the security a little I would recommend adding a basic check on the form permission would be great as it's "possible" for the template itself to contain hard coded sensitive information.

But that security check could also be a fast follow and not necessary for this to land to address our users current issues.

PR #656, opened 2025-05-06
Isolate quoting of cells for csv exports, always quote usernames

jegelstaff:
So there's a bit of reorganization in here, nothing major. Whereas we used to do some analysis of element metadata and the value that was going in the cell, in order to determine if it should be quoted, and that was all done together... this breaks it up into smaller bits, because we need just the quoting part to properly encapsulate the names of users, to cover the case where they have commas in them (or quotes as well actually).

### DIFF
--- a/modules/formulize/include/functions.php
+++ b/modules/formulize/include/functions.php
@@ -1325,18 +1325,12 @@ function prepExport($headers, $cols, $data, $fdchoice, $custdel, $template, $fid
 
         $c_uid = display($entry, 'creation_uid');
         $c_name_q = q("SELECT name, uname FROM " . $xoopsDB->prefix("users") . " WHERE uid='$c_uid'");
-        $c_name = $c_name_q[0]['name'];
-        if (!$c_name) {
-            $c_name = $c_name_q[0]['uname'];
-        }
+        $c_name = $c_name_q[0]['name'] ? $c_name_q[0]['name'] : $c_name_q[0]['uname'];
         $c_date = display($entry, 'creation_datetime');
         $m_uid = display($entry, 'mod_uid');
         if ($m_uid) {
             $m_name_q = q("SELECT name, uname FROM " . $xoopsDB->prefix("users") . " WHERE uid='$m_uid'");
-            $m_name = $m_name_q[0]['name'];
-            if (!$m_name) {
-                $m_name = $m_name_q[0]['uname'];
-            }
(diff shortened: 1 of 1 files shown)


jegelstaff:
@arvinsingla I've done a few test exports, including with usernames that have commas (in a local ESAT copy), and things seem to work OK. Doing some test exports yourself that include the username columns, and data (and usernames) with commas and quotes in them, would be helpful.

And if you want to know more about the exportIntroChar thing, check out the description of that preference option in Formulize. Crazy hacky stuff to play nice with Excel vs Google, which do not respond the same to different "hint" characters. It's all meant to help avoid the "smart" detection that the spreadsheet programs do, which can often get things wrong, so we want everything that's not explicitly a number to come in as text, and not get garbled unnecessarily by auto-conversions when opening. Because most users know zero about carefully importing csvs and declaring which data type each column has.

arvinsingla:
Just read through the code and it looks fine. Doing a functional test the exports seem correct as well. However one thing I noticed is that there's always leading spaces on all the data.

```csv
"User who made entry","Creation date",username
"	admin",2025-05-06 16:52:56,"	bob"
"	admin",2025-05-06 16:52:59,"	bob'"
"	admin",2025-05-06 16:53:04,"	""bob smith"""
"	admin",2025-05-06 16:53:13,"	Peter's Parker(2)"
"	admin",2025-05-06 16:53:20,"	Bob & Smi'th"
"	Arv'in",2025-05-06 16:58:36,"	YOooooo"
```

jegelstaff:
Yes, this is the tab leading character, which is there by default in all quoted values. This is to force Excel to treat it as text, whereas without that it would do its automatic voodoo and some values would magically become dates in Excel. This is what is controlled with the Formulize preference for the "exportIntroChar" mentioned above.

jegelstaff:
A bit more info... https://superuser.com/questions/318420/formatting-a-comma-delimited-csv-to-force-excel-to-interpret-value-as-a-string

jegelstaff:
Without this 00003456 would become 3456, for example as well.

jegelstaff:
Excel thinks its so smart

jegelstaff:
But of course, Google Sheets follows different rules and you have to use ' instead of tab. Hence the preference, because it is up to the users to figure out what is right given their standard use case. We default to: tabs because Excel is most common, except if the makecsv.php file is generating the export and then we use ' because that is more likely to be some kind of API call from Google (at least for Actua). But people can control the behaviour including turning it off completely through the prefs.

arvinsingla (approved):
LGTM 👍🏽

PR #658, opened 2025-05-10
FIX: Ensuring breadcrumbs are properly generated for all pages

arvinsingla:
* Adds a `title` variable to all the markdown pages
* Adds a modified version of the [jekyll breadcrumb plugin](https://github.com/git-no/jekyll-breadcrumbs)
* Adds `<title>` attribute to the default template
* Adds unrelated fix for mobile Safari by using a pre-processed hero background image instead of using CSS to process the blend mode


arvinsingla:
@jegelstaff it would be great if you could run it locally and go through the pages to ensure I didn’t miss anything.

jegelstaff:
Yes, that's the only thing left. Code looks good.

PR #660, opened 2025-05-13
Fix/more breadcrumbs and pages

jegelstaff:
Trying to get breadcrumbs loaded from config into github pages


jegelstaff:
This was successfully merged! GitHub glitch meant it was not closed at that time. Closing now.

PR #704, opened 2025-07-15
MCP refactor

arvinsingla:
A refactor of the current work in progress PR(#695) for adding MCP server support. This refactor accomplishes the following

1. Creates a new `FormulizeMCPException` class (`mcp/mcpException.php`)
2. Standardize all output by the MCP server to `handleHTTPRequest` method
3. Ensures that all errors throw an exception which bubbles up to `handleHTTPRequest` for final output
5. Updating all methods that previously handled output to instead either return or throw
6. Moving the `FormulizeMCP` class to it's own file `mcp/mcp.php`
7. Moving documentation endpoint to it's own standalone file `mcp/docs.php`

### DIFF
--- a/mcp/mcp.php
+++ b/mcp/mcp.php
@@ -355,16 +355,14 @@ class FormulizeMCP
 	private function handleInitialize()
 	{
 		return [
-			'result' => [
-				'protocolVersion' => '2024-11-05',
-				'capabilities' => [
-					'tools' => [],
-					'resources' => [],
-					'prompts' => []
-				],
-				'serverInfo' => $this->system_info(),
-				'instructions' => $this->getInitializeInstructions()
-			]
+			'protocolVersion' => '2024-11-05',
+			'capabilities' => [
+				'tools' => [],
+				'resources' => [],
+				'prompts' => []
+			],
+			'serverInfo' => $this->system_info(),
+			'instructions' => $this->getInitializeInstructions()
 		];
 	}
 
--- a/mcp/prompts.php
+++ b/mcp/prompts.php
@@ -99,9 +99,7 @@ trait prompts {
 	private function handlePromptsList()
 	{
 		return [
-			'result' => [
-				'prompts' => array_values($this->prompts)
-			]
+			'prompts' => array_values($this->prompts)
 		];
 	}
 
@@ -126,9 +124,7 @@ trait prompts {
 		try {
 			$messages = $this->generatePrompt($promptName, $arguments);
 			return [
-				'result' => [
-					'messages' => $messages
-				]
+				'messages' => $messages
 			];
 		} catch (Exception $e) {
 			throw new FormulizeMCPException(
--- a/mcp/resources.php
+++ b/mcp/resources.php
@@ -115,9 +115,7 @@ trait resources {
 		$this->registerResources();
 
 		return [
-			'result' => [
-				'resources' => array_values($this->resources)
-			]
+			'resources' => array_values($this->resources)
 		];
 	}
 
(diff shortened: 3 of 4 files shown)


mcp/mcpException.php
```diff
@@ -0,0 +1,126 @@
...
+			case 'form_not_found':
+				$error['troubleshooting'] = [
+					'issue' => 'Invalid form ID',
+					'solutions' => [
+						'Use list_forms tool to get valid form IDs',
+						'Verify the form exists and is accessible'
+					]
+				];
+				break;
+			case 'unknown_tool':
```
jegelstaff, line 103: This should be 'unknown element' or something like that? There was handling in the original based on yucky string searching and conventions about how the messages should be worded. Better to structure it this way with issues, etc, but I think some of the magic has been lost. Please check the original error handling code and make sure the cases where there were messages with the magic strings that were searched for, are supported here by proper issue types, etc

mcp/mcp.php
```diff
@@ -0,0 +1,560 @@
...
+		header('Access-Control-Max-Age: 0');
+
+		// Set the HTTP response code
+		http_response_code($httpResponseCode);
+
+		// Set the content type to JSON
+		if ($isJSON) {
+			header('Content-Type: application/json; charset=utf-8');
+			// Encode the body as JSON
+			echo json_encode($body, JSON_PRETTY_PRINT);
```
jegelstaff, line 108: In general, it seems that everything lower down is doing some variation on this...

```
eturn [
			'jsonrpc' => '2.0',
			'result' => [
				'prompts' => array_values($this->prompts)
			],
			'id' => $id
];
```

What I'm concerned about is the jsonrpc and result structure, and id. If everything always returns stuff with this overall basic structure, should we put that here instead, and the lower methods would return only the result part or something like that? There may be reasons I'm not aware of why that's a bad idea. I'm just thinking of it from the point of view of avoiding repetition and right now we're repeating a lot of this stuff over and over in the lower methods.

mcp/index.php
```diff
@@ -713,33 +10,24 @@ private function JSONerrorResponse($message, $code = -32603, $id = null, $contex
...
-	http_response_code(500);
-	echo json_encode([
-		'error' => 'Server initialization failed: ' . $e->getMessage(),
-		'timestamp' => date('Y-m-d H:i:s')
-	]);
+		throw new FormulizeMCPException('MCP Server is disabled', 'server_disabled');
+	}
+} catch (FormulizeMCPException $e) {
+	$server->sendResponse([
+		'jsonrpc' => '2.0',
```
jegelstaff, line 26: This is perhaps where my other comment about the jsonrpc and result and id structure, runs into trouble... because not all responses have that structure. :(

Still, if there's a way to steamline what the lower methods are returning, I think that's good?

jegelstaff (changes requested):
In general, terrific! A few comments about the formatting/composition of the response strings, and the nuanced handling of error cases.

Also, I am assuming you have run this for real and tested that it actually gives responses, etc. You haven't just written the code alone, right?

mcp/mcp.php
```diff
@@ -317,18 +322,26 @@ public function handleHTTPRequest()
...
 				throw new FormulizeMCPException(
 					'Invalid endpoint: ' . $cleanPath,
 					'method_not_found',
 					-32601,
 				);
 			}
+
+			$response = [
+				'jsonrpc' => '2.0',
+				'result' => $result,
```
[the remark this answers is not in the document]
arvinsingla: You're right. I reviewed the spec and the way I set it up would result in the double `result` issue. I've now upaded the return values to properly specify the nested value.  7c477383ee7b0b4157528b2d07706451c9725ee9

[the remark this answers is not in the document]
jegelstaff: Latest commit looks good!

jegelstaff (approved):
Looks good!

PR #723, opened 2025-08-25
Add create_form and create_form_element tools to MCP

jegelstaff:
Start with the basics. Creating forms, adding elements to forms. Editing properties and so on will come later. Editing screens, etc, will come later. Right now, creating forms simply works the same as manual creation of forms on the admin side. This is meant to allow the AI to give you a head-start, a basic set of working stuff that you can refine through the interface.

The real magic is in the element handling, because creating the linked elements will result in the connections (relationships) being instantiated, and the embedding automatically of subforms, etc. 

To handle the element creation properly, each element class needs to declare an example/instructions string that is used in the tool schema for the create_form_element tool, so the AI knows the options that are possible for each kind of element. This necessitates the creation of class files for all the elements, finally!

### DIFF
--- a/modules/formulize/class/formulize.php
+++ b/modules/formulize/class/formulize.php
@@ -32,7 +32,6 @@
 ###############################################################################
 
 include_once XOOPS_ROOT_PATH.'/modules/formulize/include/common.php';
-
 #[AllowDynamicProperties]
 class FormulizeObject extends XoopsObject {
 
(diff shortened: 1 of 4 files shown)


jegelstaff:
Getting started. Not ready for review yet.

arvinsingla:
I know this is a work in progress, but wanted to offer a question. Do we think it's worthwhile to somehow incorporate the work done by the configuration as code system here? For example leveraging the creation of forms and elements using the methods there? That way the machine just needs to build the JSON schema for what a system needs to look like and can feed that into the existing methods.

jegelstaff:
> I also wonder if a bunch of this should really just be in the insert method of the form handler? 🤔

Yup, a bunch of this is now in the insert method of the form handler. :)

modules/formulize/class/formulize.php
```diff
@@ -100,3 +99,141 @@ static function serializedDBFields() {
 	}
 
 }
+#[AllowDynamicProperties]
+class formulizeHandler {
```
jegelstaff, line 103: This is not really a handler like the other handlers. But I followed the same structure in terms of the file because that would be consistent? But the methods are just static and accessed directly.

Since they're static, they could just be methods on the formulizeObject. But they do all kinds of things beyond what would just be tools/methods pertaining to a form or an element or a screen, etc. So it seemed like they should be at a different level, so a "formulizeHandler" made sense? Or did it?

Very open to suggestions about where this should live, how this should be organized.

PR #724, opened 2025-08-28
MCP - prepare new options

jegelstaff:
This is a stub of work, aiming at allowing MCP to request creation of an entry, in which a specified option is not currently an option but new options are allowed for that element (linked selectbox autocomplete), and the system will do the creation automatically. Code is not finished, needs testing, but also semantics of how the AI would handle this is not clear, since the AI would likely create the source entry first and then create the entry that links to it, by using a foreign key explicitly. So what's the point? This approach needs further review and consideration. However, the encapsulation of the creation of new entries and cleanup of the prepDataForWrite function, is laudable all on its own.

### DIFF
--- a/mcp/tools.php
+++ b/mcp/tools.php
@@ -850,7 +850,9 @@ private function validateFilter($filter) {
 
 			// Step 2: Prepare and validate the data
 			$preparedData = [];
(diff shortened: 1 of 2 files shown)


jegelstaff:
Might be a useful reference, but definitely not merging this ever

PR #725, opened 2025-08-30
Separate class files for each element type

jegelstaff:
Finally, must do this for AI element creation tools, so each element can declare things cleanly/easily to the AI. All the other benefits will come along with this as well, especially streamlined element options, etc.

### DIFF
--- a/modules/formulize/class/data.php
+++ b/modules/formulize/class/data.php
@@ -1366,3 +1366,19 @@ class formulizeDataHandler {
 
 }
 
+/**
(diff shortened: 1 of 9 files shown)


jegelstaff:
Going to close the linked issue after all... only text for display (both kinds) and grid will be left non-classed. Everything else will have been done, and the existing classes will all be AI enabled at least at a basic level. Worth closing the issue and making more specific issues for further refinements.

jegelstaff:
@arvinsingla When tests pass, please merge at your discretion :)

PR #726, opened 2025-09-05
Tests for form and element creation

jegelstaff:


jegelstaff:
@arvinsingla See if you can get these all passing locally for you. There may be issues with the Change Columns popup in the last tests. And I have found it's sometimes flaky about clicking some stuff, depending on the timing of animations and stuff, and seems to only be able to function if the browser is visible on screen. But I have got them all passing, with some annoying timeouts where internal operations and async requests are happening.

arvinsingla:
Generally speaking with with playwright you shouldn't need to use timeouts. Instead you should be asserting based on auto retry assertions. https://playwright.dev/docs/test-assertions#auto-retrying-assertions

PR #731, opened 2025-09-16
Support conds w. { } refs, linked elmts, multi vals both sides, using LIKE

jegelstaff:
This one is pretty specific. Right side is a multi value linked list (ie: checkboxes), left side is a multi value linked list (ie: multiselect autocomplete), both link to the same source. Condition is "left LIKE right" accomplished via {handle} ref.

In this case, if the right has multiple values, we need the query to be structured like this:

(left LIKE val1 OR left LIKE val2 OR left like val3) and target_entry = entryID

Previous to this, query would end up as:

left LIKE val1,val2,val2 AND target_entry = entryID

And that query would fail completely.

If the right has a single value, then the original structure is OK. We only slip into this mode if there are multiple values on the right side.

A deeper refactor of the entire condition generate SQL would be nice. But it's so fundamental to so many operations, and works, that we're not going to touch it for now.


jegelstaff:
Two lines removed in an ammendment to master.

PR #733, opened 2025-09-18
FIX: Removing docker filesystem mount for database folder

arvinsingla:
In order to deal with case sensitivitiy issues in table names for Formulize as well as ensure that our local development environments more closely resemble production this PR removes the file system mount for the database files in the container.

Changes:
- Removal of the directory mount for mysql data
- Creation of a volume mount for the mysql data
- adding a volume to the docker compose for mariadb data
- Creating a formulize specific mariadb `cnf` that explicitely sets  `lower_case_table_names` to 0

### DIFF
--- a/docs/deploying_locally.md
+++ b/docs/deploying_locally.md
@@ -12,14 +12,7 @@ Formulize can be run locally using Docker containers. Among other things, this m
 
 ## First, if you're running Windows
 
-1. Find this file in your repository: ```/docker/maraidb/conf.d/lower_case_table_names.cnf.windows```
-
-2. Copy or rename it to: ```/docker/maraidb/conf.d/lower_case_table_names.cnf```\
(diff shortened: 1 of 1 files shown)


jegelstaff:
Looking good. Docs need updating. 

Also, investigate if there's any way that the existing data dir on host file system can take precedence, so that people don't suddenly lose their data when they start this up? Or what about a migration, where there's a script that runs on startup that imports the current database in the data dir to the volume, if the data dir exists? Make it a user runnable script. EXCEPT: what about importing from Windows and your table names have lowercase or wonky case names?? Maybe we don't want to do an import after all.

Also, what about directing somehow the container to mount different db volumes? So an archive of volumes could be in existence and you just use a different one depending which branch your container is using? Like a manual config somewhere as an extension of the main compose... something to look into.

PR #747, opened 2025-10-18
Pull jquery resources locally instead of from the internet

arvinsingla:
I was unable to run the e2e tests on the train when the internet was down because the jQuery resources were trying to load from the internet. Since the browser is fully reset between every test there's no caching of these resources.


jegelstaff (approved):
Looks good to me. On the other hand, in production systems, like a high traffic formulize.net, we would want to pull from some kind of high volume provider like Google, not dish out all that ourselves? But I don't think that's any reason not to do this change now.

arvinsingla:
Actually.... I think I want to take a different approach here and use environment variables to dictate behaviour. Default  will be to pull from the remote cdn. But if $ENVIRONMENT=local We'll use the included version.

PR #761, opened 2025-11-11
Resurrect session timeout warning for Anari

jegelstaff:
Session Timeout Warning was in the standalone theme and never made it into Anari! Should perhaps be made a core feature... but exact method of display could/should be part of theme I guess.


jegelstaff:
Shows the actual timeout time to the user. Depends on CSS that should come through because of incremented v number on stylesheet. Dependent on the theme, so revised theme would need to include this!

PR #769, opened 2025-11-15
Check that tokens is writable

jegelstaff:


jegelstaff:
- [ ] check it works when folder missing
- [ ] check message appears by itself and when op patchDB etc is triggered

PR #778, opened 2025-12-04
Dynamic default values, based on other elements

jegelstaff:
Specify that an element should use as its default value, the value entered into another element in a given entry. Filter for the given entry based on { } references to elements in the current form.


jegelstaff:
Needs to pass tests, and needs to check what happens when there is no matching element, etc. We should redo the element based on its canonical defaults or no default?

PR #789, opened 2025-12-19
Form-based user and group management

jegelstaff:
User account management should be simplified to be driven by forms and their entries. ie: a Staff form generates users from every entry, a Departments form generates groups, etc.


jegelstaff:
- [x] Add editing support when the User Account elements are present in a form
- [x] Default timezone needs to respect system default? For user account form element.
- [x] Need to get values saving cleanly and unique value checking working 
- [x] Need some control over mapping of plain elements to user elements. A referral form might have plain elements for email, etc, and when a referral meets conditions to become a regular entries are users entry, those mappings should be used to create the user record automatically, if they exist. Otherwise someone has to create the user manually, which might be OK in some situations but not ideal.
- [x] A given entries-are-users form should have standard group memberships for the users, as an optional setting, so user groups don't need to be manually assigned after users are created
- [x] add optional conditions for when a group should apply to the user, similar to the conditions for when user accounts should be available for the entry, so that you can only put people in the managers category group for example and the managers yes no box ticked off.
- [x] Fix the assign to cat group based on text... doesn't dynamically update and should say "group for the [form singluar] selected in..."
- [x] need one entry per user etc to be a per group setting so participants can have one entry per user in the participants form but other people can still create many. Current settings just get set as the default for registered users and there's always registered users files and you can just add others if necessary.
- [x] Have a setting that makes the user in entries_are_users form, owner of the entry
- [x] Need to apply template group permissions to created groups when entries are created in the form. And if the permissions for template groups are changed, entry groups need to be updated.
- [x] Add display support for including the User Account elements among the columns in the list of entries
- [x] Add API support for accessing User Account properties in datasets
- [x] 2FA only visible if 2FA is enabled in the website!
- [x] 2FA behaviours matching what edituser.php does now
- [x] EAU or EAG should be part of a triple setting at the top of a form, like multiple entries per user, etc. What kind of information is recorded in this form? Regular Data, Users, Groups ??? We might want the EAU and EAG to be less hidden... problem is that it's not just users and groups, it's data about those things, plus representing, so maybe the current UI is OK as long as it's high enough in the Settings tab??
- [x] Add separate Users and Groups front side menu section for management of all accounts and groups independent of the entries-are-users forms and entries-are-groups forms
- [x] Get entry locking working for user account elements
- [x] Touch up UI for buttons and options in entries-are-users/groups (get rid of grey gradient, etc)
- [x] Ensure specific group groupscope settings are property written AND removed from template to entry groups when writing
- [x] Add handling for existing user accounts based on Futureworx scenarios
- [x] Add UI option for toggling on all groups vs just plain and template groups, in permissions UI (for webmasters only, as a way of double checking stuff)
- [x] Provide active/inactive toggle for users with edit rights, besides the active user
- [x] Does group membership get enforced based on sdf condition on notes for Jazz2
- [x] When looking at Engage New Glasgow (entry group) how do we manage the membership of All Users vs other categories like Participants? Right now the UI is showing up with the members based on... All Users? This needs to be a lot more nuanced.
- [x] When users is turned off... everyone loses uid? Seems wrong. Check participant list. Is the uid element being removed??

jegelstaff:
@copilot Can you please review this code. I want you to focus on structural issues and code organization issues, whether there are new functions or methods that could or should be refactored together into one, to improve future maintenance of the code, for example. I want to merge this with the master branch but I want it to be as clean as possible. I don't want to merge code that is unpolished in obvious ways.

jegelstaff:
@copilot  Can you please review this code. I want you to focus on structural issues and code organization issues, whether there are new functions or methods that could or should be refactored together into one, to improve future maintenance of the code, for example. I want to merge this with the master branch but I want it to be as clean as possible. I don't want to merge code that is unpolished in obvious ways.

I previously asked for this, and one commit was made: abfa0ca4d37316407ec06196aea19ed9825ee2b0 and then there was an authentication problem, and it seems like the agent/session got stuck after that. Please do a complete review of the entire PR and all files added/changed as part of it.

PR #820, opened 2026-03-10
Deduplicate workflows, optimize couple steps.

ettavolt:
The system so far resisted most of the tricks I have up my sleeves! I've only managed to improve a couple of smaller things:

- Use built-in compose health check instead of an extra step.
- Cache Playwright binaries.
- Not recording video unless explicitly requested.

I've tried to use GitHub Container Registry, but that part of the cloud is no faster than the one where Docker Hub is. Caching "image.tar" is even slower.
I've also put DB on a tmpfs, but it does not improve the runtime a single bit.
Creating a custom image is unlikely to help, as now it is only container images that are slow to set up. And I'm sure they are loaded into a daemon outside the runner container, i.e. we can't cache them.
Unless, of course, the image is an "omnibus" with _installed_ php and RDBMS and thus doesn't use docker (-in-docker).

I guess, any slow down is now completely with Playwright. I'll see if I can figure out where exactly. Maybe it is actually Playwright itself (Selenium@JVM FTW!)

-Arseniy


docker-compose.yaml
```diff
@@ -3,22 +3,26 @@ services:
...
+        condition: service_healthy
     volumes:
       - .:/var/www/html/
       - ./trust:/var/www/trust
       - ./docker/php/conf.d/formulize.ini:/usr/local/etc/php/conf.d/formulize.ini
     ports:
-      - '8080:80'
-    links:
-      - mariadb
+      - '127.0.0.1:8080:80'
```
ettavolt: I strongly believe this should be the default binding, as there is no TLS.

If really wanted, the "local address" part could be overridden via [the "merge" feature](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/#merging-rules).

docker-compose.yaml
```diff
@@ -3,22 +3,26 @@ services:
...
-      - mariadb
+      mariadb:
+        condition: service_healthy
     volumes:
       - .:/var/www/html/
       - ./trust:/var/www/trust
       - ./docker/php/conf.d/formulize.ini:/usr/local/etc/php/conf.d/formulize.ini
     ports:
-      - '8080:80'
-    links:
```
jegelstaff: @arvinsingla Don't we nee the links: section to connect the web server to the db server over the imaginary docker internal network or whatever? I assume @ettavolt was able to successfully spin things up without this though?

.github/workflows/e2e-test-manual.yml
```diff
@@ -39,53 +46,45 @@ jobs:
         cache: 'npm'
         cache-dependency-path: tests/e2e/package-lock.json
 
-    - name: Create .htaccess file in project root
```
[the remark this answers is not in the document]
ettavolt: I should have mentioned that this is dropped only because playwright makes the same setup already, see `tests/e2e/htaccess-setup.js`.
(Which it should be, given that it should be runnable outside of GitHub workflow as well.

PR #828, opened 2026-03-27
Add/update docblocks to follow PHPDoc pattern

herbdool:
@jegelstaff I've added or updated docblocks in functions.php. It's a really overwhelming file. It started off as me wanting to use an LLM to document the functions here so I have a better idea of how things work for a client site. But even the LLM struggled to find all the functions so I had to still do some manual work. At any rate, it seems to be fairly accurate now.


jegelstaff:
Hello world... I am trying to finish this today.

jegelstaff:
Epic review and adjustments complete. Claude double checked also. Looking good, merging! Thank you for putting the first pass together!

PR #859, opened 2026-06-23
Make filters work with standard metadata fields

### LINKED ISSUE #679: Dropdown search option needs to work for users (and other metadata)
If you select Dropdown List as the search type for the creation user (or mod user) column, there's a fatal error.

jegelstaff:


jegelstaff:
Make fid pass through to the filter level, split the handling of user metadata fields off into another function, and improve the handling of the searches to cover the negative exclusion case.

PR #867, opened 2026-07-10
Adding EAU/EAG awareness to AI tools

jegelstaff:
When making entries via AI tools in a form, if it is an entries-are-users or entries-are-groups form, the correct behaviours need to occur, same as if you're using the normal UI.


jegelstaff:
Included in #888

PR #870, opened 2026-07-13
Extendable element classes

jegelstaff:
Element classes had a few subtle dependencies outside the core classes. We have removed those so they can all be extended cleanly, and determinations like "is this a radio button" will still work, when you extend the radio button class, for example.


jegelstaff:
Re-merged with #872

PR #871, opened 2026-07-13
ThemeEditor

adamak93:
Shows you themes installed.
Current ACTIVE theme selected by default.
Also shows you the folders and files of the theme.
Display the code editor textarea in the page, default selected the theme.html, 20 rows by default
add disclaimer: DONT MESS WITH ANYTHING IF YOU DONT KNOW WHAT YOURE DOING

- SIDE NOTES

modules/formulize/templates/admin/application_code.html (See below)

<textarea id="applications-custom_code"  name="applications-custom_code" class="code-textarea canValidate" rows="20"/>
<{$content.custom_code}></textarea>

don't use canValidate - causes files to be sent to the server for php syntax checks. the files we're checking are NOT php. they're primarily html css, and maaaaybe javascript.

Use this to set the theme default.
global $xoopsConfig;
$xoopsConfig['theme_set'];
 

Check mailusers.php and mailusers.html for an indication of how to approach the layout + data side stuff for the . (HTML for the layout PHP for data side)

PS: DONT COMMIT GITIGNORE AND MAINFILE.PHP! AND MAKE COMMITS ON A SEMI-REGULAR BASIS!


jegelstaff:
Thank you @adamak93 !

jegelstaff:
Pushed a commit to use specific modes for css and js, which might not make much practical difference but could help. Not tested! But it is a pretty simple change so fingers crossed I didn't break it on you.

PR #903, opened 2026-08-21
New formulize.org

jegelstaff:
This update to the website is focused on a few things, related to making the on ramp easier for everyone:

- Make the "tiers" clear and provide a way for people to easily download or sign up
- Make news/posts a first class section that can be featured/maintained easily
- Preserve all the existing documentation, videos, etc


arvinsingla:
Overall I think this is a great first step. There are a few things I think need work.

1. I find the fonts and colours hard to read. Especially transitioning between the heading font to the paragraph font.
2. Baseline font sizes I think need to go up. Feels like 14px should be the base and there should be more breathing room.
3. Both examples of the pricing strucure feel really busy to me. But the second feels better. It might help for readability of you highlighted the currently hovered row.

Was this designed with Claude Design? If not it might be worthwhile to feed it into Claude Design and have it do some refinement.

jegelstaff:
Cool, thanks. Yes, font sizes are wonky, readability is an issue. And the side to side card rail is a wall of text for sure. This was originally in Claude Design, then I revised some things in Claude code, because I wanted to be able to live iterate over it while it's in Jekyll, and now I'm going back to Claude Design since the tier model has pretty much been finalized.

Any suggestions about the level names? Boring old bronze, silver, gold, platinum? Yuck. But it's that, or basic, standard, pro, enterprise, also Yuck. Or something no one understands because you're trying to be too cute. I don't know.

